PR #63, opened 2023-07-28
EVEREST-177 Manage BackupStorages in k8s

oksana-grishchenko:
EVEREST-177 Add/Delete ObjectStorages

Summary
- use names as ID for the backup-storages, remove the ID field
- k8s objects are created when the DatabaseCluster is created
- do not allow to remove storage if any DatabaseCluster referencing it
- add the `Description` field to consist with UI design


michal-kralik (dismissed):
@oksana-grishchenko can you please elaborate a bit on what the expected context/usage of `everestK8s` is? What functionality is it expected to hold?

At the moment, we have:
- `api/backup_storage.go`
- `models/backup_storage.go`

and we're adding `api/everestK8s.go` to handle backup storage as well.

I think with the current approach the file will grow big quickly.
What about using `api/backup_storage.go` for this with the `EverestServer` struct?

michal-kralik:
@oksana-grishchenko on a side-note, I think we may be missing creation of backup storages on `RegisterKubernetesCluster`.
I believe this use case shall work:
1. Add k8s cluster
2. Add Backup storage
3. Remove k8s cluster
4. Add new k8s cluster
5. Spin up a db cluster with backup storage

Would we have the backup storage available in step 5 in k8s?

oksana-grishchenko:
@michal-kralik 
> @oksana-grishchenko can you please elaborate a bit on what the expected context/usage of everestK8s is? What functionality is it expected to hold?

`everestK8s` interface contains methods for `EverestServer` to communicate with k8s without diving into the details (like creating k8s client etc), you only need to know the `kubernetesID`. 
It's like a bridge between Everest and k8s, now it contains the proxing method and the methods for managing the ObjectStorages. That methods have additional logic like Secrets creation, but the interface user doesn't need to care about it and it's hidden inside the implementation. 

> I think with the current approach the file will grow big quickly.
If the interface will grow we could split it into sub-interfaces (one for ObjectStorages, one for MonitoringConfigs) as you did for the `storage` interface. 

So the main purpose of this interface is to hide the k8s communication details. Also such approach allows us to mock k8s interaction in unit-tests if needed. 

> and we're adding api/everestK8s.go to handle backup storage as well.
so it's not the purpose of this interface. 

> What about using api/backup_storage.go for this with the EverestServer struct?
no, this file doesn't know anything about k8s and it interacts with it via the new `everestK8s` interface. 

I would be happy to elaborate further on this topic if there are still any concerns, please let me know if any. 

---------

> on a side-note, I think we may be missing creation of backup storages on `RegisterKubernetesCluster`.

Great catch, I'll update it, thanks!

api/everestK8s.go
```diff
@@ -0,0 +1,246 @@
...
+func (e *everestK8sImpl) ApplyObjectStorage(ctx echo.Context, kubernetesID string, bs BackupStorage, secretFields map[string]string) error { //nolint:funlen
+	k, err := e.storage.GetKubernetesCluster(ctx.Request().Context(), kubernetesID)
+	if err != nil {
+		e.l.Error(err)
+		return ctx.JSON(http.StatusInternalServerError, Error{
+			Message: pointer.ToString("Could not get a Kubernetes cluster"),
+		})
+	}
+
+	everestClient, err := perconak8s.NewFromSecretsStorage(
```
gen1us2k, line 50: what about `kubernetes.NewFromSecretsStorage`? IMO it's more clear

gen1us2k (commented):
Thanks for the implementation. Overall it looks great. However, I left some comments how we can improve the implementation

api/everestK8s.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	reverseProxy.Transport = transport
+	req := ctx.Request()
+	req.URL.Path = buildProxiedURL(ctx.Request().URL.Path, kubernetesID, resourceName, cluster.Namespace)
+	reverseProxy.ServeHTTP(ctx.Response(), req)
+	return nil
+}
+
+func (e *everestK8sImpl) getDBClustersByObjectStorage(ctx context.Context, everestClient *perconak8s.Kubernetes, storageName string) ([]everestv1alpha1.DatabaseCluster, error) {
+	list, err := everestClient.ListDatabaseClusters(ctx)
```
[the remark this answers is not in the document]
recharte: This sounds like a great idea but I don't think it'll work. AFAIK you can't just use any field as a selector.

Core types have some options to use as field selectors (see [here](https://hoelz.ro/blog/which-fields-can-you-use-with-kubernetes-field-selectors)) but as far as I'm aware there's no way to that with CRDs ([you're stuck with name and namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#supported-fields)).

This has been in discussion by the community for several years, see https://github.com/kubernetes/kubernetes/issues/53459 and https://github.com/kubernetes/kubernetes/issues/1362

api-tests/tests/backup-storages.spec.ts
```diff
@@ -46,45 +47,48 @@ test('add/list/get/delete backup storage success', async ({request}) => {
...
     expect(one.ok()).toBeTruthy();
     expect((await one.json()).name).toBe(payload.name)
 
     // update
     const updatePayload = {
-        name: 'backup-storage-name1',
         bucketName: 'percona-test-backup-storage1',
         accessKey: "otherAccessKey",
         secretKey: "otherSecret"
     }
```
recharte, line 75: Please also add a description change here and assert it's being correctly applied.

recharte (commented):
Overall, great job Oksana. There are few comments that need addressing though.

api/everest.go
```diff
@@ -75,7 +75,8 @@ func (e *EverestServer) initStorages() error {
 	return err
 }
 
-func (e *EverestServer) proxyKubernetes(ctx echo.Context, kubernetesID, resourceName string) error {
+// ProxyKubernetes proxies echo requests to the k8s.
+func (e *EverestServer) ProxyKubernetes(ctx echo.Context, kubernetesID, resourceName string) error {
```
michal-kralik, line 79: `ProxyKubernetes` does not seem to be used outside this module.
Can we keep it private?

migrations/000005_update_backup_storages_table.up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE backup_storages DROP CONSTRAINT backup_storages_pkey;
+ALTER TABLE backup_storages ADD PRIMARY KEY (name);
+ALTER TABLE backup_storages DROP COLUMN id;
```
michal-kralik, line 3: I strongly encourage to keep the `id` column and rather make `name` a unique index.
What would be the reasons to remove the `id` column?

api/backup_storage.go
```diff
@@ -65,66 +67,47 @@ func (e *EverestServer) CreateBackupStorage(ctx echo.Context) error { //nolint:f
...
-					secretKeyID,
-				)
-			}
-		}
-	}()
-
-	accessKeyID = uuid.NewString()
-	err = e.secretsStorage.CreateSecret(c, accessKeyID, params.AccessKey)
-	if err != nil {
+	existingStorage, err := e.storage.GetBackupStorage(c, params.Name)
```
michal-kralik, line 71: Would it be possible to create a common logic for `CreateBackupStorage` so it can be reused to create `MonitoringConfig` with the same logic?
If you're not sure what the best approach could be, I'll be happy to look into this with my implementation.

api/backup_storage.go
```diff
@@ -65,66 +67,47 @@ func (e *EverestServer) CreateBackupStorage(ctx echo.Context) error { //nolint:f
...
 	if params.Url != nil {
 		url = *params.Url
 	}
 
+	var description string
+	if params.Description != nil {
+		description = *params.Description
+	}
+
 	s, err := e.storage.CreateBackupStorage(c, model.CreateBackupStorageParams{
```
michal-kralik, line 102: We need to make sure to revert these changes in case of an error later.

api/kubernetes.go
```diff
@@ -89,6 +87,12 @@ func (e *EverestServer) RegisterKubernetesCluster(ctx echo.Context) error {
 		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Could not store kubeconfig in secrets storage")})
 	}
 
+	err = e.applyExistingStorages(ctx, *k)
+	if err != nil {
+		// log the error but do not fail the process
+		e.l.Error(err)
```
michal-kralik, line 93: I believe we need a more advanced logic here.
Maybe not in this PR, but I suggest we:

1. First check if the resources are already created
2. Warn if any exist
3. Provide `force` option to overwrite the resources.

michal-kralik (commented):
A couple comments before submitting a full review.

gen1us2k:
>UPD: now we apply all existing BackupStorages to a registered cluster.

@michal-kralik @oksana-grishchenko I'm not sure that we really need this because IMO it has a couple of drawbacks
1. The same backup storage may not be in use on all Kubernetes clusters and this might be a security breach if one of k8s clusters were compromised. 
2. We simply do not need to copy all of them on all k8s clusters we have.

I'd like to suggest creating one during the creation of a database cluster and checking the existence of this object at this stage instead of copying everything to all k8s clusters

api/everestK8s.go
```diff
@@ -0,0 +1,246 @@
...
+	secretName := buildSecretName(bs.Name)
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: k.Namespace,
+		},
+		StringData: secretFields,
+		Type:       corev1.SecretTypeOpaque,
+	}
+	_, err = everestClient.CreateSecret(ctx.Request().Context(), secret)
```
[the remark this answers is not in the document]
oksana-grishchenko: the comment is not relevant anymore, the k8s communication to the existing `kubernetes` package and the secrets are created using kubeClient

api/everestK8s.go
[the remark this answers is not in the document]
oksana-grishchenko: ran a couple of tests, discussed it with Diogo and Michal, considered the links Diogo shared regarding the fieldSelectors for custom resources - looks that there is no way to expose the selectors outside of the k8s indeed.

pkg/kubernetes/client/client.go
```diff
@@ -95,12 +107,12 @@ func (c *Client) GetServerVersion() (*version.Info, error) {
 	return c.clientset.Discovery().ServerVersion()
 }
 
-// ListDatabaseClusters returns list of managed PCX clusters.
+// ListDatabaseClusters returns list of managed database clusters.
 func (c *Client) ListDatabaseClusters(ctx context.Context) (*everestv1alpha1.DatabaseClusterList, error) {
 	return c.dbClusterClient.DBClusters(c.namespace).List(ctx, metav1.ListOptions{})
```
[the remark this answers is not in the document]
oksana-grishchenko: https://github.com/percona/percona-everest-backend/pull/63#discussion_r1287272585

model/backup_storage_helpers.go
```diff
@@ -90,12 +89,12 @@ func (db *Database) ListBackupStorages(_ context.Context) ([]BackupStorage, erro
...
-	storage := &BackupStorage{
-		ID: id,
-	}
-	err := db.gormDB.First(storage).Error
+// GetBackupStorage returns BackupStorage record by its Name.
+func (db *Database) GetBackupStorage(_ context.Context, name string) (*BackupStorage, error) {
+	storage := &BackupStorage{}
+	// fixme: for some reason, gorm doesn't understand the Name field as a PrimaryKey,
+	// so "Where" is added as a quickfix
+	err := db.gormDB.Where("name = ?", name).First(storage).Error
```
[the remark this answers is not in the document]
oksana-grishchenko: tried different appoaches, still doesn't work without specifing field in a query. Changed the `Where` to the `First` so it looks a bit better, but I wouldn't spend more time to the problem in the scope of this PR.

pkg/kubernetes/client/objectstorage/object_storage.go
```diff
@@ -0,0 +1,138 @@
...
+
+// ObjectStorage returns a db cluster client.
+func (c *Client) ObjectStorage(namespace string) Interface { //nolint:ireturn
+	return &client{
+		restClient: c.restClient,
+		namespace:  namespace,
+	}
+}
+
+// Interface supports list, get and watch methods.
```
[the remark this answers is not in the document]
oksana-grishchenko: updated, thanks

api/backup_storage.go
```diff
@@ -266,15 +318,14 @@ func (e *EverestServer) UpdateBackupStorage(ctx echo.Context, backupStorageID st
...
 		return ctx.JSON(httpStatusCode, Error{Message: pointer.ToString(err.Error())})
 	}
 
 	e.deleteOldSecretsAfterUpdate(c, params, s)
 
 	result := BackupStorage{
-		Id:         updated.ID,
 		Type:       BackupStorageType(updated.Type),
 		Name:       updated.Name,
 		BucketName: updated.BucketName,
```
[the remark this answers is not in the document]
oksana-grishchenko: Thanks for noticing it! Updated

api/backup_storage.go
```diff
@@ -387,13 +438,26 @@ func (e *EverestServer) checkStorageAccessByUpdate(ctx context.Context, storageI
...
+	ctx context.Context, backupStorageName string, params *UpdateBackupStorageParams,
 	newAccessKeyID, newSecretKeyID *string,
 ) (*model.BackupStorage, int, error) {
 	updated, err := e.storage.UpdateBackupStorage(ctx, model.UpdateBackupStorageParams{
-		ID:          backupStorageID,
-		Name:        params.Name,
+		Name:        backupStorageName,
 		BucketName:  params.BucketName,
 		URL:         params.Url,
 		Region:      params.Region,
```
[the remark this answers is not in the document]
oksana-grishchenko: Thanks for the catch, updated

api-tests/tests/backup-storages.spec.ts
oksana-grishchenko: done, thanks

migrations/000005_update_backup_storages_table.up.sql
oksana-grishchenko: The problem was discussed in the BE team and it was considered to use the name as identifier since we don't have a clear vision for multi-tenancy now and bc one of the initial problems was that using the `ID` is confusing for the the API users who basically only need the `name` field. So it was decided to use the k8s approach - referencing resources by its name

api/backup_storage.go
oksana-grishchenko: For k8s resources the `createConfigWithSecret` is the function with such common logic. 
	
As for the API handlers, considering the fact that we moved CRs creation to the DBCluster creation time, I don't think we need a generic method for the API handlers anymore, it may bring more problems than it solves in case we would need to change a behaviour for one of the config types. 
	
The logic that creates/updates/deletes `ObjectStorages` during the `DatabaseCluster` modifications could be indeed reused for the monitoring configs, but to move things forward faster, I didn't create any universal interface for that in this PR,  feel free to create it during the monitoring implementation.

api/backup_storage.go
oksana-grishchenko: not relevant anymore since it was decided to move CRs creation to the DBCluster creation time.

api/kubernetes.go
oksana-grishchenko: not relevant anymore since it was decided to move CRs creation to the DBCluster creation time.

oksana-grishchenko:
> I'd like to suggest creating one during the creation of a database cluster and checking the existence of this object at this stage instead of copying everything to all k8s clusters

@gen1us2k from now on, the CRs creation happens during the DatabaeCluster creation

gen1us2k (approved):

recharte (approved):
Good job Oksana, it's good to go.

PR #66, opened 2023-08-01
EVEREST-261: Initial proposal for API design changes

michal-kralik:
EVEREST-261

A couple notes:
- Create/Update/Delete API endpoints returns HTTP 200 with an empty response


maxkondr:
I will try to provide an answer from general stand point, mostly rely on API usage convenience (actually the reason why API is created) and user experience. 

> > Is the performance issue that big of an issue? Not sure about this. It seems to me like a premature optimization. On the other hand we have convenience of the API users who don't have to make an additional request to check everything is created as expected. Or in case of our frontend - to update the UI.
> 
> Performance is not the only concern. There's a couple other criteria as well:

Well, right now we don't have neither measures nor complains regarding performance. Without that it looks like a  premature optimization because we can't prove that:
- performance really becomes better after such changes
- the issue with performance is more important/valuable rather than user experience, so that we may decrease the API usage convenience level for the sake of performance improvement.
 
> * The response would be used by a negligible percentage of users. 

We don't have any statistics to claim that. 

I agree that this approach (omitting full resource in a response body) may be very beneficial in case the returned body  is completely the same as request that user sent. But in our case we have an opposite situation. Looking at `CreateDatabaseCluster` object schema:
- there is only one mandatory part of parameters - `engine`. So it can be quite common situation that user sets mandatory param only.
- in order to obtain the exact values that server decided to use for the rest of fields in `DBCluster` object user needs to send an additional request. 

This complicates the user experience since he has to do an extra actions in order to achieve his goal. 
On another hand - in case user doesn't need the response (such situations really take place) - he just ignores the body. No extra actions are required for that from user side.

> * You can always issue another GET request if you need the response

An additional request is ALWAY more expensive than returning the body (from user and server stand point):

1. user shall initiate a new request:
1.1 build request
1.2 initiate a new connection with TLS handshake. Yes, it is possible to re-use the existing connection, but this is not so widely used practise so far.
1.3 send request

2. server shall process the request:
2.1 authentication
2.1 authorization
2.3 validation
2.4 fetching the resource to return
2.5 send response

So if we are talking about general performance optimisation, we have:
- 1.1 - 1.3 + 2.1 - 2.5 actions in case of additional `GET` request
vs
- 2.4 - 2.5 actions for case of returning the resource in response.

> * Returning a response may require additional code which creates room for errors

there are actions to mitigate it: code review, tests, ...

> I understand it may look as a no-brainer at first sight to return the response, but after analyzing the use cases and the percentage of users affected, it becomes more clear this brings only marginal benefits.

Right now we try to stick to general standards/guidelines/best practises for building API and product in general. 

Returning the new object fits with the REST principle of "Uniform Interface - Manipulation of resources through representations." The complete object is the representation of the new state of the object that was created.

There is a really excellent reference for API design, here: [Best Practices for Designing a Pragmatic RESTful API](http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api)

To sum up:
- I agree that this approach may take place
- It decreases the level of API usage conviniese
- We need to have a strong evidences that is it worth being applied here right now.

Later the situation may change and there will be a better time for considering it.
For now - let's postpone with adoption.

PR #72, opened 2023-08-04
EVEREST-173: resource usage endpoint

michal-kralik:
- Returns available and allocated resources per cluster
- Resource allocation functionality has been ported from PMM


gen1us2k (changes requested):
Thanks for the implementation. However, it misses API tests

gen1us2k (approved):

michal-kralik:
The test shall pass after https://github.com/percona/percona-everest-cli/pull/72 is merged

pkg/kubernetes/kubernetes.go
```diff
@@ -75,7 +91,21 @@ func (k *Kubernetes) ClusterName() string {
...
+	if err != nil {
+		return ClusterTypeUnknown, err
+	}
+	for _, storageClass := range storageClasses.Items {
+		if strings.Contains(storageClass.Provisioner, "aws") {
+			return ClusterTypeEKS, nil
+		}
+		if strings.Contains(storageClass.Provisioner, "minikube") ||
+			strings.Contains(storageClass.Provisioner, "kubevirt.io/hostpath-provisioner") ||
+			strings.Contains(storageClass.Provisioner, "standard") {
```
recharte, line 106: I don't think this is necessarily correct but we can fix this when we support other k8s cluster types.

recharte (approved):
I'm not really sure that this custom logic for each k8s cluster type is entirely correct but it was working for PMM and I think it serves our current needs.

pkg/kubernetes/kubernetes.go
michal-kralik: Me neither. I agree this may need a review.

PR #85, opened 2023-08-08
EVEREST-107 Fixed dev-latest build

gen1us2k:
- Fixed not found the issue for the refresh webpage
- Small refactoring after graceful shutdown implementation


api/everest.go
```diff
@@ -103,37 +107,55 @@ func (e *EverestServer) initKubeClient(ctx echo.Context, kubernetesID string) (*
...
-	httpServer.GET("/*", echo.WrapHandler(staticFilesHandler))
+	indexFS := echo.MustSubFS(public.Index, "dist")
+	// FIXME: Ideally it should be redirected to /everest/ and FE app should be served using this endpoint.
+	//
+	// We tried to do this with Fabio and FE app requires the following changes to be implemented:
+	// 1. Add basePath configuration for react router
+	// 2. Add apiUrl configuration for FE app
+	//
+	// Once it'll be implemented we can serve FE app on /everest/ location
+	e.echo.FileFS("/*", "index.html", indexFS)
```
michal-kralik, line 129: Can we instead of `/*` path use `/index.html` and `/`?

michal-kralik: What about favicon and asset manifest?

michal-kralik (dismissed):

api/everest.go
gen1us2k: No, we can't use `/index.html` or `/` because react app has its own routes (something around 8-9). So it will be a fallback for react routes. I added favicon and assets manifest

oksana-grishchenko (approved):

recharte (approved):

PR #87, opened 2023-08-08
EVEREST-107 Fixed security issues for api-tests

gen1us2k:


recharte (approved):

api-tests/package.json
```diff
@@ -24,6 +24,8 @@
   },
   "dependencies": {
     "@playwright/test": "^1.35.0",
-    "shelljs": "^0.8.5"
+    "semver": "^7.5.4",
+    "shelljs": "^0.8.5",
+    "word-wrap": "^1.2.5"
```
michal-kralik, line 29: Why do we need the new dependencies?
Anything broken with the api tests? They don't seem to be used.

michal-kralik (changes requested):

api-tests/package.json
michal-kralik: Since they're indirect, the only way to update them is to update the affected package.
I believe updating them this way makes no difference, since the affected module will use its own vulnerable version.

oksana-grishchenko (approved):
LGTM, the CI is not happy about this update though.

PR #98, opened 2023-08-09
EVEREST-233 Added backup endpoints implementation

gen1us2k:


docs/spec/openapi.yml
```diff
@@ -750,6 +752,218 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+    put:
+      tags:
+        - databaseClusterBackup
+      summary: Replace the specified cluster backup on the specified kubernetes cluster
+      description: Replace the specified cluster backup on the specified kubernetes cluster
+      operationId: updateDatabaseClusterBackup
+      parameters:
```
oksana-grishchenko, line 884: do we need the update functionality? what is the use case? I was under impression if we need another backup, we create a new one. It only has two fields - the `DBClusterName` and the `ObjectStorageName`. I understand we could provide a method just to edit the k8s CR, but it's not clear how it should work on the operator side.

docs/spec/openapi.yml
```diff
@@ -750,6 +752,218 @@ paths:
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/DatabaseClusterBackup'
+    delete:
+      tags:
+        - databaseClusterBackup
+      summary: Delete the specified cluster backup on the specified kubernetes cluster
+      description: Delete the specified cluster backup on the specified kubernetes cluster
+      operationId: deleteDatabaseClusterBackup
+      parameters:
```
oksana-grishchenko, line 935: if we need the data also deleted from the storage, which I believe is what we want to achieve, perhaps we'll need an additional effort on the operator's side to achieve this.

gen1us2k: It can be handled by upstream operators via finalizers. We can just set them. Right now we do not concern this in terms of the nearest milestone. It can be addressed later. However we have options that are easy to adopt

docs/spec/openapi.yml
gen1us2k: it will trigger reconciliation loop and change the upstream backup cr object

oksana-grishchenko: That's not the use case. I mean, what is the point of editing it at all? This is how the upstream operator's work

> You can edit statuses but once the backup started/finished/failed it doesn’t make sense to edit spec fields. Once backup is completed, the backup data won’t affected from any modifications of CR.

so I don't think why the end user would need it.

gen1us2k: removed this endpoint. Thanks for catching this

oksana-grishchenko (approved):

recharte (approved):

PR #100, opened 2023-08-10
EVEREST-270 Return 500 instead of 502 if k8s cluster is unavailable

gen1us2k:


oksana-grishchenko (approved):

public/dist/index.html
```diff
@@ -1 +1,20 @@
 
+  <!DOCTYPE html>
```
michal-kralik, line 2: Why do we need this file?

api/proxy.go
```diff
@@ -0,0 +1,105 @@
...
+		e.l.Error(err)
+		return ctx.JSON(http.StatusBadRequest, Error{
+			Message: pointer.ToString("Could not create REST transport"),
+		})
+	}
+	reverseProxy.Transport = transport
+	reverseProxy.ErrorHandler = errorHandler
+	req := ctx.Request()
+	req.URL.Path = buildProxiedURL(ctx.Request().URL.Path, kubernetesID, resourceName, cluster.Namespace)
+	req.Header.Set(everestKubernetesHeader, cluster.Name)
```
michal-kralik, line 78: Why do we set the header?

api/proxy.go
```diff
@@ -0,0 +1,105 @@
...
+	uri = strings.ReplaceAll(uri, "-", "")
+	return fmt.Sprintf("/apis/everest.percona.com/v1alpha1/namespaces/%s%s%s", namespace, uri, resourceName)
+}
+
+func errorHandler(res http.ResponseWriter, req *http.Request, err error) {
+	if errors.Is(err, syscall.ECONNREFUSED) {
+		clusterName := req.Header.Get(everestKubernetesHeader)
+		errorMessage := fmt.Sprintf("%s kubernetes cluster is unavailable", clusterName)
+		res.WriteHeader(http.StatusBadRequest)
+		b, _ := json.Marshal(Error{Message: pointer.ToString(errorMessage)}) //nolint:errchkjson
```
michal-kralik, line 100: Let's check the error please

api/proxy.go
```diff
@@ -0,0 +1,105 @@
...
+	// remove kebab-case
+	uri = strings.ReplaceAll(uri, "-", "")
+	return fmt.Sprintf("/apis/everest.percona.com/v1alpha1/namespaces/%s%s%s", namespace, uri, resourceName)
+}
+
+func errorHandler(res http.ResponseWriter, req *http.Request, err error) {
+	if errors.Is(err, syscall.ECONNREFUSED) {
+		clusterName := req.Header.Get(everestKubernetesHeader)
+		errorMessage := fmt.Sprintf("%s kubernetes cluster is unavailable", clusterName)
+		res.WriteHeader(http.StatusBadRequest)
```
michal-kralik, line 99: We shall use a different status code. This is not a bad a request since we could not connect to k8s cluster.

api/proxy.go
```diff
@@ -0,0 +1,105 @@
...
+	// cut the resource name if present
+	uri = strings.TrimSuffix(uri, resourceName)
+
+	// remove kebab-case
+	uri = strings.ReplaceAll(uri, "-", "")
+	return fmt.Sprintf("/apis/everest.percona.com/v1alpha1/namespaces/%s%s%s", namespace, uri, resourceName)
+}
+
+func errorHandler(res http.ResponseWriter, req *http.Request, err error) {
+	if errors.Is(err, syscall.ECONNREFUSED) {
```
michal-kralik, line 96: Let's please include other reasons for errors, such as timeouts, etc.

api/proxy.go
```diff
@@ -0,0 +1,105 @@
...
+		clusterName := req.Header.Get(everestKubernetesHeader)
+		errorMessage := fmt.Sprintf("%s kubernetes cluster is unavailable", clusterName)
+		res.WriteHeader(http.StatusBadRequest)
+		b, _ := json.Marshal(Error{Message: pointer.ToString(errorMessage)}) //nolint:errchkjson
+		res.Write(b)                                                         //nolint:errcheck,gosec
```
michal-kralik, line 101: Let's please check the error

api/proxy.go
```diff
@@ -0,0 +1,105 @@
...
+func buildProxiedURL(uri, kubernetesID, resourceName, namespace string) string {
+	// cut the /kubernetes part
+	uri = strings.TrimPrefix(uri, "/v1/kubernetes/"+kubernetesID)
+
+	// cut the resource name if present
+	uri = strings.TrimSuffix(uri, resourceName)
+
+	// remove kebab-case
+	uri = strings.ReplaceAll(uri, "-", "")
+	return fmt.Sprintf("/apis/everest.percona.com/v1alpha1/namespaces/%s%s%s", namespace, uri, resourceName)
```
michal-kralik, line 92: We don't properly sanitize the url components. This opens us to url attacks. Let's please sanitize the values.

api/proxy.go
```diff
@@ -0,0 +1,105 @@
...
+func errorHandler(res http.ResponseWriter, req *http.Request, err error) {
+	if errors.Is(err, syscall.ECONNREFUSED) {
+		clusterName := req.Header.Get(everestKubernetesHeader)
+		errorMessage := fmt.Sprintf("%s kubernetes cluster is unavailable", clusterName)
+		res.WriteHeader(http.StatusBadRequest)
+		b, _ := json.Marshal(Error{Message: pointer.ToString(errorMessage)}) //nolint:errchkjson
+		res.Write(b)                                                         //nolint:errcheck,gosec
+	}
+	// Keeping default behavior of error handler
+	res.WriteHeader(http.StatusBadGateway)
```
michal-kralik, line 104: Since we're handling errors, let's please return a meaningful default error.

michal-kralik (changes requested):

api/proxy.go
gen1us2k: So we can get the name from the header for a kubernetes cluster.

michal-kralik: there's no benefit in storing it in the header

gen1us2k: What alternatives do you see?

api/proxy.go
gen1us2k: namespace comes from the db and we do sanitization before storing it. As for other params we have this covered by oapi

michal-kralik: Can you please point me to the part of code which sanitizes `resourceName`? I could not find it.

api/proxy.go
michal-kralik: In case we need the cluster name for any purposes, we shall pass it to the function which requires it instead of modifying the headers. In this case it would be `errorHandler`.

api/proxy.go
```diff
@@ -93,13 +95,25 @@ func buildProxiedURL(uri, kubernetesID, resourceName, namespace string) string {
...
 func errorHandler(res http.ResponseWriter, req *http.Request, err error) {
-	if errors.Is(err, syscall.ECONNREFUSED) {
-		clusterName := req.Header.Get(everestKubernetesHeader)
-		errorMessage := fmt.Sprintf("%s kubernetes cluster is unavailable", clusterName)
-		res.WriteHeader(http.StatusBadRequest)
-		b, _ := json.Marshal(Error{Message: pointer.ToString(errorMessage)}) //nolint:errchkjson
-		res.Write(b)                                                         //nolint:errcheck,gosec
+	logger := logger.MustInitLogger()
+
+	defer logger.Sync()
```
michal-kralik, line 100: We already have the logger initialized, why do we need to initialize it again?

api/proxy.go
```diff
@@ -93,13 +95,25 @@ func buildProxiedURL(uri, kubernetesID, resourceName, namespace string) string {
...
+	clusterName := req.Header.Get(everestKubernetesHeader)
+	errorMessage := fmt.Sprintf("%s kubernetes cluster is unavailable", clusterName)
+	b, err := json.Marshal(Error{Message: pointer.ToString(errorMessage)})
+	if err != nil {
+		logger.Error(err.Error())
+	}
+	if errors.Is(err, syscall.ECONNREFUSED) || errors.Is(err, os.ErrDeadlineExceeded) {
```
michal-kralik, line 107: This will never be true, because you're checking an error from `json.Marshal`.

michal-kralik (changes requested):

api/proxy.go
gen1us2k: because net/http.Proxy does not know anything about it

gen1us2k: We can use fmt.Println instead

api/proxy.go
gen1us2k: I don't want to implement a custom proxy just to store it using variable names and make it thread-safe. Storing the k8s name in headers is just the simplest solution we can have right now because proxy.ErrorHandler and other structs have their own signature we need to follow. As another solution, we can pass the name to a function that returns the handler but I'm not sure that it will work reliably keeping in mind multik8s support

gen1us2k: So a benefit of using headers is that it's a thread-safe, simple solution without any wrappers around the standard and implemented stuff we use.

api/proxy.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	return func(res http.ResponseWriter, req *http.Request, err error) {
+		if errors.Is(err, syscall.ECONNREFUSED) || errors.Is(err, os.ErrDeadlineExceeded) {
+			res.WriteHeader(http.StatusInternalServerError)
+			if _, err := res.Write(b); err != nil {
+				logger.Error(err.Error())
+			}
+			return
+		}
+		res.WriteHeader(http.StatusBadGateway)
```
michal-kralik, line 106: As mentioned in the original comment, let's please handle the default error in a consistent manner by returning `{message:""}` json object with a meaningful error message.

api/proxy.go
michal-kralik: The middleware validates if the `name` parameter is present. It does not perform any sanitization.
As far as I know it does not check if the resource name is not for example `../name`
Happy to explain how this can be misused over Slack if needed.

michal-kralik (changes requested):

api/proxy.go
gen1us2k: It returns it

api/proxy.go
gen1us2k: This is out of the scope of this PR. I'll create a new ticket for that

api/proxy.go
gen1us2k: line 107

recharte: @gen1us2k it seems to me that in line 107 we'll return "%s kubernetes cluster is unavailable" just like in line 101. Am I missing something?

gen1us2k: https://github.com/percona/percona-everest-backend/blob/657580afe828def96fd3d01fa7988d974ce7abb6/api/proxy.go#L106-L109

recharte: Yes, but what's the content of `b`? It seems to me like it will always be "%s kubernetes cluster is unavailable" regardless of returning 500 (line 101) or 502 (line 107)

gen1us2k: Yep. Because when errorHandler is being called it means that k8s cluster is unavailable anyway

recharte: In that case why use 500 for `syscall.ECONNREFUSED` and `os.ErrDeadlineExceeded` and 502 for everything else?

gen1us2k: FE handles 500 only afaik and the webpage hangs and do not inform the end-user on 502 error

api/proxy.go
michal-kralik: Please link it here for reference

michal-kralik (dismissed):
Let's please resolve the pending conversations about a security issue ticket and default error being handled properly.

api/proxy.go
gen1us2k: Let's see it in the future. Right now the issue with 502

gen1us2k: It returns 500 in every scenario. I wanted to return 502 for some other unhandled cases to keep the default behavior of HttpSingleHostProxy

PR #101, opened 2023-08-10
EVEREST-204 Edit BackupStorages

oksana-grishchenko:
EVEREST-204 

**What's new**

When a Backup storage is updated, the related CRs (ObjectStorages, Secrets) are updated in k8s too.


gen1us2k (approved):

api/backup_storage.go
```diff
@@ -247,11 +248,24 @@ func (e *EverestServer) UpdateBackupStorage(ctx echo.Context, backupStorageName
 		return ctx.JSON(http.StatusInternalServerError, Error{Message: pointer.ToString(err.Error())})
 	}
 
+	tx := e.storage.Begin(c)
```
michal-kralik, line 251: We're creating a tx but not using it later. The `tx.Rollback()` will have no effect.

api/backup_storage.go
```diff
@@ -396,3 +410,30 @@ func (e *EverestServer) updateBackupStorage(
...
+		everestClient, err := kubernetes.NewFromSecretsStorage(c, e.secretsStorage, k.ID, k.Namespace, e.l)
+		if err != nil {
+			return errors.Wrapf(err, "could not create Kubernetes client for %s", k.Name)
+		}
+
+		err = everestClient.UpdateObjectStorage(c, k.Namespace, bs, secretData)
+		if err != nil {
+			return errors.Wrapf(err, "could not update ObjectStorage %s", bs.Name)
+		}
+	}
```
michal-kralik, line 436: We shall reconsider this logic.
In case one of the k8s cluster updates fail, we end up in an inconsistent state, where some clusters have been updated, while others were not.

michal-kralik (changes requested):

api/backup_storage.go
oksana-grishchenko: @michal-kralik  what do you mean? `tx.Rollback()` is called to rollback the changes in case the k8s objects are failed to be updated, the `tx.Commit()` is called if the changes were applied successfully.

api/backup_storage.go
oksana-grishchenko: The multi-k8s support for BackupStorages is gonna be resolved in the scope of EVEREST-257. I agree the storages needs to be consistent across all k8s clusters for all methods we have.

michal-kralik: Since it's not part of this this ticket, can we remove this functionality? In this state, I believe it's broken and will cause issues.

api/backup_storage.go
michal-kralik: That's true, however in order to execute a query within a transaction, you need to call methods on the `tx` object.
See https://gorm.io/docs/transactions.html for more details.
Happy to jump on a call to explain in more detail if needed.

michal-kralik (changes requested):
We shall address outstanding conversations.

api/backup_storage.go
oksana-grishchenko: The approach however works for a single k8s cluster that we support now. So to address the concern regarding inconsistent state in case of an error for multi k8s, the method now throws an error in case there are several k8s clusters, so we can keep the value for single cluster and prevent inconsistence for multiple. Wdyt?

michal-kralik: I think it's more consistent with the single k8s cluster approach we have.

michal-kralik (approved):
Looks great, thank you

PR #102, opened 2023-08-11
EVEREST-248 Make kubernetes error messages userfriendly

gen1us2k:
https://jira.percona.com/browse/EVEREST-248

PR based on the changes in https://github.com/percona/percona-everest-backend/pull/100


michal-kralik (changes requested):
Let's please address comments in #100 first

api/proxy.go
```diff
@@ -86,6 +109,51 @@ func buildProxiedURL(uri, kubernetesID, resourceName, namespace string) string {
 	return fmt.Sprintf("/apis/everest.percona.com/v1alpha1/namespaces/%s%s%s", namespace, uri, resourceName)
 }
 
+func everestModifier(resp *http.Response) error {
```
recharte, line 112: I think we should rename this function to something more meaningful, it should allude to the modification it does which is this case is to change the status message if we face an error.
Something like
```suggestion
func mapStatusMessageCRDsOnError(resp *http.Response) error {
```

oksana-grishchenko (approved):
Looks great! 
I thought it would be great to have unit-tests for the error messaging, which will also serve as a documentation about how it works, but it might change the scope, so leaving it up to you.

michal-kralik (approved):

PR #110, opened 2023-08-14
EVEREST-182 List backups endpoint

gen1us2k:


fabio-silva:
@gen1us2k just to be sure, will this endpoint return:
- backup name;
- backup size;
- backup type
?

gen1us2k:
@fabio-silva backup.name, state and that's all. No size at this moment and no backup.type at this moment as well

oksana-grishchenko (approved):
The API looks a bit confusing bc we use the
`/kubernetes/{kubernetes-id}/database-clusters/{name}/backups` pattern only for listing the backups, while the other actions are still under the
`/kubernetes/{kubernetes-id}/database-cluster-backups` pattern, but for the sake of not blowing up the scope - I'm fine with it.

gen1us2k:
@oksana-grishchenko From the product standpoint we need to list only per cluster backups and do not need to have list-backups that we had before. Listing backups can be solved without this endpoint using the old design for it but this will be easier for our FE team and API consumers to work with. That's why I implemented this way

api/database_cluster_backup.go
```diff
@@ -16,10 +16,32 @@
...
+	"github.com/AlekSi/pointer"
+	"github.com/labstack/echo/v4"
+)
 
 // ListDatabaseClusterBackups returns list of the created database cluster backups on the specified kubernetes cluster.
-func (e *EverestServer) ListDatabaseClusterBackups(ctx echo.Context, kubernetesID string) error {
+func (e *EverestServer) ListDatabaseClusterBackups(ctx echo.Context, kubernetesID string, name string) error {
+	req := ctx.Request()
+	if !validateRFC1123(name) {
+		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("cluster name is not RFC 1123 compatible")})
```
michal-kralik, line 33:
```suggestion
		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Cluster name is not RFC 1123 compatible")})
```

michal-kralik (approved):

PR #111, opened 2023-08-15
EVEREST-269 Renamed ObjectStorage to BackupStorage

gen1us2k:


michal-kralik:
I'd like to do this in my PR for monitoring if possible.
Otherwise there will be tons of conflicts and I'll have hard time merging this.

api/database_cluster.go
```diff
@@ -60,11 +60,11 @@ func (e *EverestServer) ListDatabaseClusters(ctx echo.Context, kubernetesID stri
 
 // DeleteDatabaseCluster Create a database cluster on the specified kubernetes cluster.
 func (e *EverestServer) DeleteDatabaseCluster(ctx echo.Context, kubernetesID string, name string) error {
-	err := e.deleteObjectStorages(ctx.Request().Context(), kubernetesID, name)
+	err := e.deleteBackupStorages(ctx.Request().Context(), kubernetesID, name)
```
oksana-grishchenko, line 63: I'm concerned about the naming, which becomes confusing when there are both k8s resource and Everest object have exactly the same name. 
For example here - when a DBCluster is deleted, we try to delete the related storage CRs from k8s. We don't touch the Everest object here, however from this naming I would assume we delete everything. So to understand what's happening we need to dive into details each time.

what about having a prefix like `k8s`? So we'll have `backupStorages` for Everest objects and `k8sBackupStorages` for k8s objects.

gen1us2k: I use use k8s prefix for k8s related objects

oksana-grishchenko: I see, I can find the `updateK8SBackupStorages` method, which is what I would like to have. But why we don't have the prefix in the `deleteBackupStorages` and `createBackupStorages` methods? That methods also operates with k8s objects only.

gen1us2k:
@michal-kralik Unfortunately no because the issue is on me and I made everything ready to merge.

oksana-grishchenko (approved):

michal-kralik:
> @michal-kralik Unfortunately no because the issue is on me and I made everything ready to merge.

I can make the changes in my PR.
It's going to be tons of work if we merge this first instead of me making the changes in the other PR which will further postpone the delivery.

recharte:
> 

@michal-kralik As discussed offline, this issue is blocking the FE development for the backups and that's a higher priority now compared to the monitoring part. Let's not block this due to the monitoring PR.
Sorry for the extra work you'll have to do fixing the conflicts.

pkg/kubernetes/client/object_storage.go
recharte, line 1: You should also rename this file.

public/dist/index.html
```diff
@@ -1 +1,20 @@
...
+      try { window.__REACT_DEVTOOLS_GLOBAL_HOOK__ = window.parent.__REACT_DEVTOOLS_GLOBAL_HOOK__; } catch {}
+      </script>
+      <!-- minimal css resets -->
+      <style> body { margin: 0; } </style>
+    <link rel="icon" href="/favicon.ico"><script defer src="/static/main.ca9a2bd539386a10e771.js"></script></head>
+    <body>
+      <div id="root"></div>
+    </body>
+  </html>
+
```
recharte, line 20: Why is this part of this PR? Was it intentional?

recharte (changes requested):

public/dist/index.html
gen1us2k: Nah. I just test be with fe locally and it makes these changes

PR #112, opened 2023-08-15
EVEREST-209: monitor k8s cluster

michal-kralik:
EVEREST-209
EVEREST-164
EVEREST-186

- Add support for monitoring k8s cluster
- Moved `VMAgent` deployment to the backend
- Support disabling k8s cluster monitoring via API
- Introduced generic `EnsureConfigExists` for both monitoring config and backup storage
- Does not support multiple k8s clusters
- Does not update `MonitoringConfig` in k8s on `MonitoringInstance` update
- Does not delete monitoring config on monitoring instance deletion
- Upgrade to Go 1.21
- Migrate old `pmm_instances` table but does not support reverting the operation
- Copy a couple Kubernetes client `*Object()` methods from `cli`
- Rename "pmm instance" to "monitoring instance"
- Tests pass locally
- Requires https://github.com/percona/percona-everest-cli/pull/98


pkg/kubernetes/configs.go
```diff
@@ -67,42 +106,39 @@ func (k *Kubernetes) DeleteObjectStorage(ctx context.Context, name, namespace st
...
 	}
 	_, err := k.CreateSecret(ctx, secret)
 	// if such Secret is already present in k8s - consider it as created and do nothing (fixme)
-	if err != nil && !k8serr.IsAlreadyExists(err) {
+	if err != nil && !k8serrors.IsAlreadyExists(err) {
 		return err
 	}
 
-	err = apply(secretName, namespace)
+	err = k.client.CreateResource(ctx, cfg, &metav1.CreateOptions{})
```
gen1us2k, line 138: ApplyObject does the same if you want to create resources without thinking about the type. However, the preferred way is to call the typed function that creates corresponding object type in k8s

pkg/kubernetes/client/customresouces/resource.go
```diff
@@ -0,0 +1,125 @@
...
+func (c *Client) objectResource(obj runtime.Object) (schema.GroupVersionResource, error) {
+	gvk, err := c.objectKind(obj)
+	if err != nil {
+		return schema.GroupVersionResource{}, err
+	}
+
+	return c.restMapper.ResourceFor(gvk.GroupVersion().WithResource(gvk.Kind))
+}
+
+// ListResources returns a list of k8s resources.
```
gen1us2k, line 62: listResources is unused. Why do we need to have it?

api/kubernetes.go
```diff
@@ -194,6 +196,101 @@ func (e *EverestServer) GetKubernetesClusterResources(ctx echo.Context, kubernet
...
+		})
+	}
+
+	if mi == nil {
+		return ctx.JSON(http.StatusBadRequest, Error{
+			Message: pointer.ToString("Could not find the provided monitoring instance by name"),
+		})
+	}
+
+	if err := kubeClient.EnsureConfigExists(ctx.Request().Context(), mi, e.secretsStorage.GetSecret); err != nil {
```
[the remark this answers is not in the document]
michal-kralik: How do we do this?
Do we want to add management of k8s clusters where a specific monitoring instance can be used?

api/monitoring_instance.go
```diff
@@ -0,0 +1,263 @@
...
+
+	return ctx.JSON(http.StatusOK, e.monitoringInstanceToAPIJson(i))
+}
+
+// UpdateMonitoringInstance updates a monitoring instance based on the provided fields.
+func (e *EverestServer) UpdateMonitoringInstance(ctx echo.Context, name string) error {
+	params, err := validateUpdateMonitoringInstanceRequest(ctx)
+	if err != nil {
+		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString(err.Error())})
+	}
```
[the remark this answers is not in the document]
michal-kralik: The update and delete operations do not touch the `MonitoringConfig` in k8s - but they should.
It's not clear how to approach this given there can be multiple k8s clusters registered with Everest.

We have the same situation with backup storages.
I think we need to come to an agreement how to deal with this.
I updated the PR comment so it's more clear. I forgot about it when writing the description.

api/monitoring_instance.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+		e.l.Error(err)
+		return ctx.JSON(http.StatusInternalServerError, Error{
+			Message: pointer.ToString("Could not get monitoring instance"),
+		})
+	}
+	if i == nil {
+		return ctx.JSON(http.StatusNotFound, Error{
+			Message: pointer.ToString("Could not find monitoring instance"),
+		})
```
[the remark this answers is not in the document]
michal-kralik: Yes, that's a good point. See discussion in update operation for details.

pkg/kubernetes/client/customresouces/resource.go
```diff
@@ -0,0 +1,125 @@
...
+) error {
+	gvr, err := c.objectResource(into)
+	if err != nil {
+		return err
+	}
+
+	err = c.restClient.
+		Get().
+		Namespace(namespace).
+		Resource(gvr.Resource).
```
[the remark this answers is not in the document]
michal-kralik: I looked at how we would rewrite the `EnsureConfigExists` method using generics and I think it will become more complicated and harder to use/maintain. 
I can give it a try, but I don't see an issue with this approach since it uses the same client as other custom resources.

pkg/kubernetes/client/customresouces/resource.go
michal-kralik: Originally I used it for listing VMAgents. Now it's not needed. Removed

api/backup_storage.go
```diff
@@ -412,14 +412,14 @@ func (e *EverestServer) updateBackupStorage(
 	return updated, 0, nil
 }
 
-func (e *EverestServer) updateObjectStorages(c context.Context, bs model.BackupStorage) error {
-	secretData, err := e.getStorageSecrets(c, bs)
+func (e *EverestServer) updateObjectStorages(ctx context.Context, bs model.BackupStorage) error {
+	secretData, err := bs.Secrets(ctx, e.secretsStorage.GetSecret)
```
gen1us2k, line 416: Why do we need to have these changes that are unrelated to monitoring integration?

api/database_cluster.go
```diff
@@ -141,58 +146,32 @@ func (e *EverestServer) createBackupStorages(c context.Context, kubernetesID str
...
+	kubeClient, err := kubernetes.NewFromSecretsStorage(
 		c, e.secretsStorage, k.ID, k.Namespace, e.l)
 	if err != nil {
 		return errors.Wrap(err, "Could not create k8s client")
 	}
 
 	processed := make([]string, 0, len(names))
 	for name := range names {
-		err = e.createObjectStorage(c, everestClient, name, k.Namespace)
+		bs, err := e.storage.GetBackupStorage(c, name)
```
gen1us2k, line 157: Why do we need these changes?

api/database_cluster.go
```diff
@@ -141,58 +146,32 @@ func (e *EverestServer) createBackupStorages(c context.Context, kubernetesID str
...
-		return errors.Wrap(err, "Failed to get secret from secrets storage")
-	}
-
-	// create the storage and the related secrets
-	err = everestClient.CreateObjectStorage(c, namespace, *bstorage, secrets)
-	if err != nil {
-		return errors.Wrap(err, "Could not create ObjectStorage")
+		err = kubeClient.EnsureConfigExists(c, bs, e.secretsStorage.GetSecret)
+		if err != nil {
+			e.rollbackCreatedObjectStorages(c, processed, kubeClient)
```
gen1us2k, line 164: Do we really need to rollbackChanges considering that we do not have clear requirements for handling unavailable k8s and there can be only 404 or 400 errors. IMO right now instead of doing something smart it would be better to return meaningful error to the user with some action item

api/database_cluster.go
```diff
@@ -277,49 +262,39 @@ func (e *EverestServer) updateBackupStorages(c context.Context, kubernetesID, db
...
 	for name := range toCreate {
-		err = e.createObjectStorage(c, everestClient, name, k.Namespace)
+		bs, err := e.storage.GetBackupStorage(c, name)
 		if err != nil {
-			e.rollbackCreatedObjectStorages(c, processed, everestClient, k.Namespace)
-			return errors.Wrapf(err, "Could not create CRs for %s", name)
+			return errors.Wrap(err, "Could not get backup storage")
+		}
+
+		err = kubeClient.EnsureConfigExists(c, bs, e.secretsStorage.GetSecret)
```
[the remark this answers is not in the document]
gen1us2k: Can we have it in this PR please?

api/database_cluster.go
```diff
@@ -277,49 +262,39 @@ func (e *EverestServer) updateBackupStorages(c context.Context, kubernetesID, db
...
+		if err != nil {
+			e.rollbackCreatedObjectStorages(c, processed, kubeClient)
```
gen1us2k, line 272: Do we need to rollbackChanges?

api/database_cluster.go
```diff
@@ -277,49 +262,39 @@ func (e *EverestServer) updateBackupStorages(c context.Context, kubernetesID, db
...
 	}
 
 	processed = make([]string, 0, len(toDelete))
 	for name := range toDelete {
-		err = e.deleteObjectStorage(c, everestClient, name, k.Namespace, &oldCluster.Name)
+		err = e.deleteObjectStorage(c, kubeClient, name, &oldCluster.Name)
 		if err != nil {
-			e.rollbackDeletedBackupStorages(c, processed, everestClient, k.Namespace)
-			return errors.Wrapf(err, "Could not delete CRs for %s", name)
+			e.rollbackDeletedBackupStorages(c, processed, kubeClient)
```
gen1us2k, line 290: Do we need to rollback changes?

api/kubernetes.go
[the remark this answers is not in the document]
gen1us2k: The main problem right now during unprovisioning that it tries to remove VM agent installations for unmonitored clusters. I thought since we get some information from API to check the cluster exists it would be great to understand that monitoring for this k8s cluster is enabled.

api/monitoring_instance.go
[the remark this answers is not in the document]
gen1us2k: So this means that it does not solve the problem of database clusters being monitored at least for now. Since we agreed that we support only single k8s right now it would be great to have this enabled. Everest monitors k8s clusters right now and there's no way to have this monitoring integration for database clusters. Please add those changes

api/monitoring_instance.go
```diff
@@ -0,0 +1,263 @@
...
+	params *UpdateMonitoringInstanceJSONRequestBody,
+) error {
+	err := e.storage.UpdateMonitoringInstance(name, model.UpdateMonitoringInstanceParams{
+		Type:           (*model.MonitoringInstanceType)(&params.Type),
+		URL:            &params.Url,
+		APIKeySecretID: apiKeyID,
+	})
+	if err != nil {
+		go func() {
+			_, err := e.secretsStorage.DeleteSecret(ctx.Request().Context(), *apiKeyID)
```
gen1us2k, line 235: This is dangerous because the user won't see this error

api/monitoring_instance.go
```diff
@@ -0,0 +1,263 @@
...
+				e.l.Warnf("Could not delete secret %s from secret storage due to error: %s", apiKeyID, err)
+			}
+		}()
+
+		e.l.Error(err)
+		return ctx.JSON(http.StatusInternalServerError, Error{Message: pointer.ToString("Could not update monitoring instance")})
+	}
+
+	if apiKeyID != nil {
+		go func() {
```
gen1us2k, line 246: Please make it synchronous

model/backup_storage_helpers.go
```diff
@@ -61,6 +66,46 @@ type BackupStorage struct {
 	UpdatedAt time.Time
 }
 
+// SecretName returns the name of the k8s secret as referenced by the k8s MonitoringConfig resource.
+func (b *BackupStorage) SecretName() string {
+	return fmt.Sprintf("%s-secret", b.Name)
+}
+
+// Secrets returns all monitoring instance secrets from secrets storage.
+func (b *BackupStorage) Secrets(ctx context.Context, getSecret func(ctx context.Context, id string) (string, error)) (map[string]string, error) {
```
gen1us2k, line 75: Why do we need those changes and how they are related to monitoring integration?

model/backup_storage_helpers.go
```diff
@@ -61,6 +66,46 @@ type BackupStorage struct {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to get accessKey")
+	}
+	return map[string]string{
+		b.SecretKeyID: secretKey,
+		b.AccessKeyID: accessKey,
+	}, nil
+}
+
+// K8sResource returns a resource which shall be created when storing this struct in Kubernetes.
```
gen1us2k, line 90: Why do we need those changes and how they are related to monitoring integration?

pkg/kubernetes/client/customresouces/resource.go
```diff
@@ -0,0 +1,104 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+func (c *Client) objectKind(obj runtime.Object) (schema.GroupVersionKind, error) {
```
gen1us2k, line 29: Why do we need this client?

pkg/kubernetes/configs.go
```diff
@@ -8,56 +8,95 @@ import (
...
+		return errors.Wrap(err, "could not get Kubernetes resource object")
+	}
+
+	acc := meta.NewAccessor()
+	name, err := acc.Name(config)
+	if err != nil {
+		return errors.Wrap(err, "could not get name from a config object")
+	}
+
+	r, err := cfg.K8sResource(k.namespace)
```
gen1us2k, line 51: Why do we need to call this method twice?

pkg/kubernetes/configs.go
```diff
@@ -8,56 +8,95 @@ import (
...
+	if err != nil {
+		return errors.Wrap(err, "could not get Kubernetes resource object")
+	}
+
+	err = k.client.GetResource(ctx, name, r, &metav1.GetOptions{})
```
gen1us2k, line 56: Why can't we use GetMonitoringConfig here?

pkg/kubernetes/monitoring_config.go
```diff
@@ -0,0 +1,154 @@
...
+	inUse, err := k.isMonitoringConfigUsedByVMAgent(ctx, name)
+	if err != nil {
+		return false, err
+	}
+
+	if inUse {
+		return true, nil
+	}
+
+	dbs, err := k.ListDatabaseClusters(ctx)
```
gen1us2k, line 81: It's not an optimal way of listing database clusters. We can use [labelSelectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) here and check by using len(dbs.Items) and Kubernetes will do the filtration on their side

api/monitoring_instance.go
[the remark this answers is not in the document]
oksana-grishchenko: Should we focus on a single k8s cluster for now? I mean, we could use the same approach we use for BackupStorages so we would be at least sure that the data is consistent for the different DatabaseClusters inside one k8s cluster.

api/database_cluster.go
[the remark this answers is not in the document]
michal-kralik: It's a separate feature/ticket. I prefer separating the PRs not having to increase the size of this PR.

api/validation.go
```diff
@@ -186,8 +186,8 @@ func validateCreateBackupStorageRequest(ctx echo.Context) (*CreateBackupStorageP
 	return &params, nil
 }
 
-func validateCreatePMMInstanceRequest(ctx echo.Context) (*CreatePMMInstanceJSONRequestBody, error) {
-	var params CreatePMMInstanceJSONRequestBody
+func validateCreateMonitoringInstanceRequest(ctx echo.Context) (*CreateMonitoringInstanceJSONRequestBody, error) {
+	var params CreateMonitoringInstanceJSONRequestBody
```
oksana-grishchenko, line 190: since we're gonna use the "name" field as CR name, we need to check if it is RFC-1123 compatible, check the`validateRFC1123` method

api/backup_storage.go
michal-kralik: They're closely related.
Both backup storage and monitoring config share the same logic when it comes to CRUD operations on the config.

Both operate with two k8s resources:
- The configuration itself
- A related secret

This approach allows us to share the logic for CRUD, transactions, rollbacks, etc. so we don't have to maintain it in multiple places.
From this perspective this is a great candidate for a common interface.

api/monitoring_instance.go
recharte: @gen1us2k can you elaborate on why you find this change is needed? Let's try to have clear and constructive comments and not orders.

api/database_cluster.go
michal-kralik: The backup storage struct is required in the next steps. We retrieve secrets, k8s resources, etc.

api/database_cluster.go
michal-kralik: There was no change to the rollbacks. This is how it's originally implemented.
This method handles a single cluster scenario only. I believe it's appropriate.

api/database_cluster.go
michal-kralik: There was no change to the rollbacks. This is how it's originally implemented.
Not sure if it's required here though. Maybe @oksana-grishchenko can jump in?

api/database_cluster.go
michal-kralik: There was no change to the rollbacks. This is how it's originally implemented.
Not sure if it's required here though. Maybe @oksana-grishchenko can jump in?

api/monitoring_instance.go
michal-kralik: I agree. We need to figure out how to approach this. We use the same approach in other places as well.
We could consider a state machine to queue this type of actions.

model/backup_storage_helpers.go
michal-kralik: This provides a common interface for all CRUD logic regarding the configs. 
I shared more info in https://github.com/percona/percona-everest-backend/pull/112#discussion_r1295770909.
Let's continue the discussion there.

api/backup_storage.go
gen1us2k: But we have an additional ticket for further improvements for backup storage related objects specifically and the design may change. Can we have this reverted?

pkg/kubernetes/client/customresouces/resource.go
michal-kralik: This supports a common interface for all CRUD logic regarding the configs. 
I shared more info in https://github.com/percona/percona-everest-backend/pull/112#discussion_r1295770909.
Let's continue the discussion there.

api/monitoring_instance.go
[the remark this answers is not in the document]
gen1us2k: However, this PR has at least 2 tickets to solve. I'd prefer to have a full feature to be merged yet it's up to you.

pkg/kubernetes/configs.go
michal-kralik: There's a call to `k.client.GetResource()`  on the following lines which ensures the config does not already exist.
We pass in a separate struct to the `GetResource()` call so that the original `config` struct stays intact.

api/monitoring_instance.go
gen1us2k: I didn't intend to make orders sorry for that. 
https://github.com/percona/percona-everest-backend/pull/112/files#r1295697875 The original comment is here. This is a dangerous way of cleaning everything up because the end user won't see any error during the secret deletion and won't know anything about it which may lead to an inconsistent state here.

pkg/kubernetes/configs.go
michal-kralik: The goal of this call is to figure out if a resource exists or not.
Knowing the type of the resource is not necessary at this point and allows us to share the CRUD logic across different config types.

pkg/kubernetes/monitoring_config.go
michal-kralik: Yes, this could be a good way to optimize the performance.
As discussed previously, we want to support the user to use e.g. `kubectl` to run their own clusters and manage them in Everest afterwards. We would have to figure out how to check if these clusters (which may not have labels) use or don't use a monitoring config.

pkg/kubernetes/client/customresouces/monitoringconfigs.go
```diff
@@ -0,0 +1,115 @@
...
+		VersionedParams(&opts, scheme.ParameterCodec).
+		Do(ctx).Into(result)
+	return result, err
+}
+
+// Update creates a resource.
+func (c *monitoringConfigClient) Update(
+	ctx context.Context,
+	storage *everestv1alpha1.MonitoringConfig,
+	pt types.PatchType,
```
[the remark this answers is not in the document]
michal-kralik: This was a copy & paste from other custom resources. Nevertheless, it's not used so I removed it.

api/database_cluster.go
```diff
@@ -205,67 +184,73 @@ func (e *EverestServer) deleteObjectStorages(c context.Context, kubernetesID str
...
-	err := everestClient.DeleteObjectStorage(c, name, namespace, exceptCluster)
+	bs, err := e.storage.GetBackupStorage(ctx, name)
+	if err != nil {
+		e.l.Error(err)
+	}
+
+	secretName := ""
+	if err == nil {
+		secretName = bs.SecretName()
+	}
```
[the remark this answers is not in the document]
michal-kralik: You're right. Fixed

api/monitoring_instance.go
```diff
@@ -0,0 +1,263 @@
...
+
+		go func() {
+			_, err := e.secretsStorage.DeleteSecret(context.Background(), i.APIKeySecretID)
+			if err != nil {
+				e.l.Warn(errors.Wrapf(err, "could not delete monitoring instance API key secret %s", i.APIKeySecretID))
+			}
+		}()
+
+		return nil
+	})
```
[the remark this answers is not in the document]
michal-kralik: Yes, that's correct. I left a comment in the function so the logic of monitoring config removal can be implemented there.
I'm not clear myself how to approach this since it's related to the multi k8s discussion we have.

api/validation.go
michal-kralik: Good catch. I totally missed this. Fixed.

pkg/kubernetes/monitoring_config.go
gen1us2k: Having labels set on the operator side solves this issue for both components be and the operator and it'll be done automatically on the operator side. However, additional changes are required on the operator side for this

api/monitoring_instance.go
gen1us2k: Why do we need to have a state machine for an operation that should be synchronous instead of reverting these operations and making then synchronous?

api/monitoring_instance.go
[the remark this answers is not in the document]
michal-kralik: Understood, thank you. Let's do it in a separate PR.

api/monitoring_instance.go
michal-kralik: The main reason is there is nothing else to be reverted.
The flow is as follows:
1. We create a new secret
2. We update the db object

The update in step 2 fails and we want to revert step 1.
Reverting step 1 fails. Both the original db object and secret are in a consistent state at this point.

State machine would allow us to retry the removal until secrets storage is available again.

pkg/kubernetes/monitoring_config.go
michal-kralik: Yes, I agree this would help with having consistent labels.
But then what would be the benefit of having labels if all `DatabaseCluster` resources would have the labels?

model/monitoring_instance.go
recharte, line 1: Did you mix up the file names? For the BackupStorages you named the file which contains the `Secrets` and `K8sResource` functions `backup_storage_helpers.go` and here you named it `monitoring_instance.go` while the `monitoring_instance_helpers.go` has the CRUD DB functions.

recharte (commented):
Overall great job Michal, this is a huge PR with lots of changes, I still need some time to have a better look at some sections but here's some initial feedback.

api/database_cluster.go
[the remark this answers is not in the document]
gen1us2k: Okay. Makes sense. Let's do this

api/database_cluster.go
gen1us2k: The rollback actually does not do anything and may harm here. We have two reasons for getting an error from k8s cluster
1. 4xx error meaning that the object still exists and k8s failed to remove it because of configuration/validation/permission/any other reason
2. We have connection timeout, 5xx or any other error and in that case the object still exists in the cluster
Since it exists it will fail with either conflict or duplicate error or something else

api/monitoring_instance.go
gen1us2k: Yep. But we are discussing different problems. I mean that the user won't see the error that is handled in the goroutine and you're discussing consistency stuff. As for now let's not use goroutines to delete/cleanup everything because we do not manage the state of goroutine and we can't control that it will succeed the better way is to have this implemented in a synchronous way. The benefits of running this asynchronously are unclear and do not add something in terms of performance because it's a 10ms max operation and we do not have any requirements/problems with the performance of communicating on the db level

pkg/kubernetes/configs.go
gen1us2k: Why do we need to keep the original struct intact?

pkg/kubernetes/monitoring_config.go
gen1us2k: Yep. Please create a ticket and let's implement labels in the next iterations

pkg/pmm/apikey.go
```diff
@@ -0,0 +1,81 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+// CreatePMMApiKey creates a new API key in PMM by using the provided username and password.
+func CreatePMMApiKey(ctx context.Context, hostname, apiKeyName, user, password string) (string, error) {
```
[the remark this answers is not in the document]
gen1us2k: We didn't change the way we have pmm integration. We had this discussed with our IT team and they proposed to generate a new secret automatically so it will be independent from key rotation or if credentials of one who provisioned this feature were revoked and other stuff.

api/database_cluster.go
michal-kralik: To me the logic seems fine.
We try to create configs and if we fail, we revert them not to keep configs/secrets in k8s.
If the revert fails, we log an error, nothing else. Ideally we would continue retrying though.

api/monitoring_instance.go
michal-kralik: Let's say this would be done in a synchronous manner. How would you suggest we handle an error when we try to delete the secret?

pkg/kubernetes/configs.go
michal-kralik: From the documentation it's not clear what happens to the provided object when `GetResource()` returns an error.
I wanted to make sure the original `config` object is not overwritten by accident.
By creating another copy we prevent any unwanted behavior. We could also do `config.DeepCopy()`. It would have the same effect.

If you're confident the `rest.Result` method `Into()` does not touch the provided object in case of an error, we don't need to call the function again.

pkg/kubernetes/client/customresouces/resource.go
```diff
@@ -0,0 +1,104 @@
...
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	if gvk.Kind != "" {
+		if !strings.HasSuffix(gvk.Kind, "List") {
+			return gvk, nil
+		}
+		gvk.Kind = strings.TrimSuffix(gvk.Kind, "List")
+		if scheme.Scheme.Recognizes(gvk) {
+			return gvk, nil
+		}
```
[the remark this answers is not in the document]
michal-kralik: This is how Kubernetes handles this internally as well. There's still some room for error, but since Kubernetes uses it, this seems safe.

pkg/kubernetes/monitoring_config.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if err := k.client.DeleteMonitoringConfig(ctx, name); err != nil {
+		return errors.Wrap(err, "could not delete monitoring config")
+	}
+
+	if secretName == "" {
+		return nil
+	}
+
+	return k.DeleteSecret(ctx, secretName, k.namespace)
```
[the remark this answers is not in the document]
michal-kralik: Once we resolve the multi k8s discussion I wanted to consider if it makes sense to have a common logic for deletes and updates. I expect it to be the case, but we'll see.

I envision the `DeleteXxxConfig()` to handle removal of all resources in k8s related to the config.
Maybe it would be better to pass the model via `ConfigK8sResourcer` and keep the logic in this package such as `kubernetes.DeleteConfig(ctx, monitoringInstance)`

We already import `model` in `kubernetes` so we're not going to run into a cyclic dependency.

pkg/pmm/apikey.go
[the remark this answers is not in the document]
michal-kralik: I understand your concern.
Originally this was in the cli so we don't run into firewall issues.

To work around this a bit I implemented two ways to add a monitoring instance:
1. You can use username/password and the backend retrieves the API key for you
2. You can pass the API key directly so the backend does not have to communicate with PMM

If we consider web UI and how monitoring instances are added, either backend has access to PMM or UI can offer API key option only because I don't think there's a way for the UI to talk to PMM.

migrations/000006_create_monitoring_instances_table.down copy.sql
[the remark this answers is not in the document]
michal-kralik: It wasn't. Thank you for pointing this out.
As noted in the PR description, I'm not restoring `pmm_instances` in `.down.sql` intentionally.

This could be an issue if we had production users, but since I don't expect any user is going to revert migrations at this stage I assumed this is fine at the moment.

api/validation.go
```diff
@@ -186,32 +186,71 @@ func validateCreateBackupStorageRequest(ctx echo.Context) (*CreateBackupStorageP
...
+	}
+
+	switch params.Type {
+	case MonitoringInstanceCreateParamsTypePmm:
+		if params.Pmm == nil {
+			return nil, errors.Errorf("pmm key is required for type %s", params.Type)
+		}
+
+		if params.Pmm.ApiKey == "" && params.Pmm.User == "" && params.Pmm.Password == "" {
+			return nil, errors.New("one of pmm.apiKey, pmm.user or pmm.password fields is required")
```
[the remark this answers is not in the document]
michal-kralik: It's either username/password or the api key - see my comment in https://github.com/percona/percona-everest-backend/pull/112#discussion_r1296173139

model/monitoring_instance.go
michal-kralik: It's a leftover from the original implementation. Ideally I'd like to separate the model into `backup_storage.go` but didn't want to make the change in this PR due to possible conflicts from other PRs.

api/monitoring_instance.go
gen1us2k: return the error to the user with 500 error

michal-kralik: I'm sorry Andrew, if you share only a couple words I can't understand your suggestion.
What will the error message say in this case? What do we expect the user to do after we show the error message? Can the regular user actually do anything about a secret not being deleted?

gen1us2k: >Can the regular user actually do anything about a secret not being deleted?

They can do something manually because they have access to Postgres and clean it manually.

>What will the error message say in this case? 

Something meaningful and clear I guess. E.g. `cannot delete secret for this monitoring config.`

> What do we expect the user to do after we show the error message?

No idea but at least they will know that something went wrong in that case

michal-kralik: The error message "cannot delete secret for this monitoring config" is misleading.
The root cause of the error is we could not update a monitoring instance. An issue with removal of an unused secret is only a side-effect which Everest shall handle automatically.

From this perspective I think the user is better off getting an error about not being able to update the monitoring instance so the user can try again.

api/kubernetes.go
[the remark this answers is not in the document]
michal-kralik: From Andrew via Slack for reference:
> you can move it to other ticket and ignore it for now

pkg/pmm/apikey.go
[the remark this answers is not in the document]
recharte: My expectation from the slack thread about this topic was that the UI would give us the API key directly and we wouldn't need to query the PMM server for it.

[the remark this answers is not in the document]
recharte: This is fine, now we have 2 options, the user can either supply the key or we can generate it.

michal-kralik:
This is ready for approvals.
The tests will not be successful before https://github.com/percona/percona-everest-cli/pull/98 and vice-versa.

We need to merge both and then figure out any test failures.

api/monitoring_instance.go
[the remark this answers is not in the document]
oksana-grishchenko: Should we delete the monitoring config at least from the current cluster? So far it looks like we're not gonna make it work properly for a single k8s cluster (which we should support now) bc we don't know how we want the future multi k8s feature work. 

About the transaction - I mean, the `tx` should be used in the storage call `e.storage.DeleteMonitoringInstance(name)` to take effect, regardless of if we have the CRs deletion after it or not. Otherwise there is no point of using a transaction at all, because the Commit and Rollback happens over the `tx` under the hood which should be used instead of the regular `db.gormDB` in the database calls.

pkg/kubernetes/configs.go
```diff
@@ -8,56 +8,95 @@ import (
...
+		if k8serrors.IsNotFound(err) {
 			return nil
 		}
 		return errors.Wrapf(err, "Failed to get ObjectStorage %s", bs.Name)
 	}
-	return k.updateConfigWithSecret(ctx, bs.Name, namespace, secretData, func(secretName, namespace string) error {
-		storage.Spec.Type = everestv1alpha1.ObjectStorageType(bs.Type)
+
+	return k.updateConfigWithSecret(ctx, bs.SecretName(), namespace, secretData, func(secretName, namespace string) error {
+		storage.Spec.Type = everestv1alpha1.BackupStorageType(bs.Type)
```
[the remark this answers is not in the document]
oksana-grishchenko: The multi k8s discussion is not gonna be resolved now, so I personally would focus on delivering a proper solution at least for a single k8s cluster. 
This refactoring is not something that blocks us currently, so it's up to you. I'm just concerned that the different approaches we use for the creation and the update make the codebase more complicated and confusing.

api/monitoring_instance.go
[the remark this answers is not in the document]
michal-kralik: About the `tx` - yes, good point.
Re removal - originally I wanted to clear at least a single cluster but the API call is generic and it's not clear which cluster had the config.

What we could do at least is to try to remove it from all k8s clusters and just ignore errors. I think that will be a better behavior than now. Would you mind if this was done in a separate PR?

recharte (approved):
:rocket:

api/monitoring_instance.go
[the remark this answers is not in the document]
oksana-grishchenko: Ok, sure, let's merge this one

oksana-grishchenko (approved):

PR #113, opened 2023-08-17
EVEREST-277 validate names and namespace

gen1us2k:
https://jira.percona.com/browse/EVEREST-277

Instead of sanitization, I decided to have validation calls. This one is not generic and before that, I tried to implement either swagger validation or using middleware.
Both other ways require a big effort or API redesign. It turned out that validation middleware or oapi-codegen does not support `pattern` definition in schema.

There are other solutions like this https://github.com/deepmap/oapi-codegen/issues/227#issuecomment-1431333836 or others in the thread but the best thing is to use StrictHandler but it requires a whole API redesign.

I'm open to other options and feel free to reject this PR. However, it solves the problem of security.

proxyKubernetes uses three variables:

1. namespace (added validation before writing the value to the database)
2. name (added validation before proxying request)
3.  resourceName (we have mapping on the be side so they are safe)

### DIFF
--- a/public/dist/index.html
+++ b/public/dist/index.html
@@ -1,1 +1,20 @@
 
+  <!DOCTYPE html>
+  <html lang="en">
+    <head>
+<title >Percona Everest</title>
+      <meta charset="utf-8">
+      <meta name="viewport" content="width=device-width, initial-scale=1.0">
+      
+      <script>
+      // Allow to use react dev-tools inside the examples
+      try { window.__REACT_DEVTOOLS_GLOBAL_HOOK__ = window.parent.__REACT_DEVTOOLS_GLOBAL_HOOK__; } catch {}
+      </script>
+      <!-- minimal css resets -->
+      <style> body { margin: 0; } </style>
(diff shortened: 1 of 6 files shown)


michal-kralik:
This is good because we validate early.
A bulletproof solution is to update the `buildProxiedURL` in https://github.com/percona/percona-everest-backend/blob/befd60bfbdc982d3f1b91830828bf40934504574/api/proxy.go#L110 to use https://pkg.go.dev/net/url#PathEscape for individual url path segments.

This will protect us from this type of attack in general.

public/dist/index.html
```diff
@@ -1 +1,20 @@
...
+      try { window.__REACT_DEVTOOLS_GLOBAL_HOOK__ = window.parent.__REACT_DEVTOOLS_GLOBAL_HOOK__; } catch {}
+      </script>
+      <!-- minimal css resets -->
+      <style> body { margin: 0; } </style>
+    <link rel="icon" href="/favicon.ico"><script defer src="/static/main.b2c7d428373610dd8875.js"></script></head>
+    <body>
+      <div id="root"></div>
+    </body>
+  </html>
+
```
oksana-grishchenko, line 20: should it be here?

PR #119, opened 2023-08-21
EVEREST-209: update delete config

michal-kralik:
EVEREST-209
EVEREST-144
EVEREST-210
EVEREST-166
EVEREST-167
EVEREST-280

- Updates / deletes config objects across all k8s clusters when working with objects in DB or database clusters
- Refactored both monitoring config and backup storage config logic
- In order to move forward with multi k8s setups, we update/delete configs across all k8s clusters
  - When an action on a specific k8s cluster fails (e.g. due to a network timeout), we log the error, but do not return an error
  - This can lead to data inconsistency across k8s clusters which needs to be eventually resolved
  - However it's better than being stuck with updates and deletes and not being able to take any action


oksana-grishchenko (approved):
Great job, Michal! I like this multi-k8s friendly approach, in that sense our codebase becomes more reliable than it was before and it makes it easier to migrate to multi k8s in the future. 
Left a couple of comments

pkg/kubernetes/backup_storage.go
```diff
@@ -0,0 +1,43 @@
...
+		if _, ok := names[name]; ok {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// BackupStorageNamesFromDBCluster returns a map of backup storage names used by a DB cluster.
+func BackupStorageNamesFromDBCluster(db *everestv1alpha1.DatabaseCluster) map[string]struct{} {
```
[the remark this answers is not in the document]
michal-kralik: I noticed these but figured they both work with different types.

`backupStorageNamesFrom` works with `api.DatabaseCluster`.
`BackupStorageNamesFromDBCluster` works with `*everestv1alpha1.DatabaseCluster`.

I'm not sure if it's expected the `api.DatabaseCluster` and `everestv1alpha1.DatabaseCluster` have a different structure - mainly in terms of pointers - but I did not want to touch it at this point since it wasn't super clear to me.

pkg/kubernetes/backup_storage.go
```diff
@@ -0,0 +1,43 @@
...
+	names := make(map[string]struct{})
+	if db.Spec.DataSource != nil && db.Spec.DataSource.BackupStorageName != "" {
+		names[db.Spec.DataSource.BackupStorageName] = struct{}{}
+	}
```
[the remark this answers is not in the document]
michal-kralik: It's not required here since the `Spec` is not a pointer.
See https://github.com/percona/percona-everest-backend/pull/119#discussion_r1301127185 for more context.

pkg/kubernetes/backup_storage.go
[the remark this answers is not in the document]
oksana-grishchenko: I see, let's keep them both then

api/database_cluster.go
```diff
@@ -59,14 +80,35 @@ func (e *EverestServer) ListDatabaseClusters(ctx echo.Context, kubernetesID stri
 
 // DeleteDatabaseCluster Create a database cluster on the specified kubernetes cluster.
```
recharte, line 81:
```suggestion
// DeleteDatabaseCluster Deletes a database cluster on the specified kubernetes cluster.
```

api/database_cluster.go
```diff
@@ -59,14 +80,35 @@ func (e *EverestServer) ListDatabaseClusters(ctx echo.Context, kubernetesID stri
...
-	return e.proxyKubernetes(ctx, kubernetesID, name)
+
+	proxyErr := e.proxyKubernetes(ctx, kubernetesID, name)
+	if proxyErr != nil {
+		return proxyErr
+	}
+
+	if ctx.Response().Status >= http.StatusMultipleChoices {
+		return nil
+	}
```
recharte, line 103: Return nil if response code >= 300?

api/database_cluster.go
```diff
@@ -84,26 +126,52 @@ func (e *EverestServer) UpdateDatabaseCluster(ctx echo.Context, kubernetesID str
...
+
+	proxyErr := e.proxyKubernetes(ctx, kubernetesID, name)
+	if proxyErr != nil {
+		return proxyErr
+	}
+
+	if ctx.Response().Status >= http.StatusMultipleChoices {
+		return nil
+	}
```
recharte, line 164: Return nil if response code >= 300?

api/database_cluster.go
```diff
@@ -136,160 +204,203 @@ func (e *EverestServer) GetDatabaseClusterCredentials(ctx echo.Context, kubernet
...
+		return errors.Wrap(err, "could not delete config in Kubernetes")
 	}
 
+	return nil
+}
+
+func (e *EverestServer) createBackupStoragesOnUpdate(
+	ctx context.Context,
+	kubeClient *kubernetes.Kubernetes,
+	oldDB *everestv1alpha1.DatabaseCluster,
```
recharte, line 307: I feel like these create/delete BackupStoragesOnUpdate functions would be clearer if they receive the `oldNames` directly instead of the the `oldDB`. They have no use for it for the `oldDB` and mixing these two types of data is confusing IMO.

I think that having a call such as `e.deleteBackupStoragesOnUpdate(context.Background(), kubeClient, oldBackupNames, newBackupNames)` is easier to understand.

api/monitoring_instance.go
```diff
@@ -225,21 +233,47 @@ func (e *EverestServer) performMonitoringInstanceUpdate(
 	ctx echo.Context, name string, apiKeyID *string, previousAPIKeyID string,
 	params *UpdateMonitoringInstanceJSONRequestBody,
 ) error {
-	err := e.storage.UpdateMonitoringInstance(name, model.UpdateMonitoringInstanceParams{
-		Type:           (*model.MonitoringInstanceType)(&params.Type),
-		URL:            &params.Url,
-		APIKeySecretID: apiKeyID,
+	var i *model.MonitoringInstance
```
recharte, line 236: Can we name this something more meaningful? Its hard to evaluate what some function calls receive when we have variables such as `i`.
```
configs.UpdateConfigInAllK8sClusters(
			context.Background(), ks, i,
			e.secretsStorage.GetSecret, e.initKubeClient, e.l,
		)
```

recharte (commented):
Great job Michal, just a couple of questions

api/monitoring_instance.go
michal-kralik: Good feedback, renamed.

api/database_cluster.go
michal-kralik: Makes sense, thank you

api/database_cluster.go
michal-kralik: As discussed, this is a great candidate for an inline comment - added.

api/database_cluster.go
michal-kralik: As discussed, this is a great candidate for an inline comment - added.

api/monitoring_instance.go
```diff
@@ -225,21 +233,47 @@ func (e *EverestServer) performMonitoringInstanceUpdate(
...
+
+		ks, err := e.storage.ListKubernetesClusters(ctx.Request().Context())
+		if err != nil {
+			return errors.Wrap(err, "Could not list Kubernetes clusters")
+		}
+
+		go configs.UpdateConfigInAllK8sClusters(
+			context.Background(), ks, i,
+			e.secretsStorage.GetSecret, e.initKubeClient, e.l,
+		)
```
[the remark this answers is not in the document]
michal-kralik: As discussed, this is a valid point and we need to address this in the future. As we agreed the suggested approach is better than the current situation and we'll focus on resolving this in the upcoming PRs.

recharte (approved):
:penguin:

oksana-grishchenko (approved):

PR #121, opened 2023-08-22
EVEREST-207: support unregistering unavailable k8s

michal-kralik:
EVEREST-207


api/kubernetes.go
```diff
@@ -119,13 +119,13 @@ func (e *EverestServer) UnregisterKubernetesCluster(ctx echo.Context, kubernetes
...
-	if err != nil {
+	_, kubeClient, code, err := e.initKubeClient(ctx, kubernetesID)
+	if err != nil && !params.IgnoreKubernetesUnavailable {
 		return ctx.JSON(code, Error{Message: pointer.ToString(err.Error())})
 	}
 
-	if params.Force == nil || !*params.Force {
-		clusters, err := client.ListDatabaseClusters(ctx.Request().Context())
+	if kubeClient != nil && !params.Force {
+		clusters, err := kubeClient.ListDatabaseClusters(ctx.Request().Context())
```
oksana-grishchenko, line 128: Looks like in case when the `params.Force` is `true` the `kubeClient` will be crated regardless there is no need for it.

oksana-grishchenko (approved):

api/kubernetes.go
michal-kralik: you're right, fixed.

api/kubernetes.go
```diff
@@ -119,13 +119,18 @@ func (e *EverestServer) UnregisterKubernetesCluster(ctx echo.Context, kubernetes
...
+		_, kubeClient, code, err = e.initKubeClient(ctx, kubernetesID)
+		if err != nil && !params.IgnoreKubernetesUnavailable {
+			return ctx.JSON(code, Error{Message: pointer.ToString(err.Error())})
+		}
 	}
 
-	if params.Force == nil || !*params.Force {
-		clusters, err := client.ListDatabaseClusters(ctx.Request().Context())
+	if kubeClient != nil && !params.Force {
+		clusters, err := kubeClient.ListDatabaseClusters(ctx.Request().Context())
```
oksana-grishchenko, line 133: My first thought was to move the `kubeClient` creation block inside this block since it's only needed there. But it's not a big deal I believe how exactly we organize it.

oksana-grishchenko (approved):

api/kubernetes.go
michal-kralik: I wanted to do it but linters complain it's nested too deep :)

recharte (approved):

PR #123, opened 2023-08-22
EVEREST-299: create backup storage config on backup

michal-kralik:
EVEREST-299

- On demand backup requires a config to be present


api/database_cluster_backup.go
```diff
@@ -47,12 +48,68 @@ func (e *EverestServer) ListDatabaseClusterBackups(ctx echo.Context, kubernetesI
...
+	// We check if the response was successful to continue with cleanup.
+	if ctx.Response().Status >= http.StatusMultipleChoices {
+		return nil
+	}
+
+	if backup.Spec.BackupStorageName != "" {
+		bsNames := map[string]struct{}{
+			backup.Spec.BackupStorageName: {},
+		}
+		go e.deleteK8SBackupStorages(context.Background(), kubeClient, bsNames)
```
recharte, line 109: Can we delete the Backup Storage? What about other DatabaseClusterBackups that use this storage?

recharte: Nevermind, I just saw you also updated `IsBackupStorageConfigInUse` to check if DBClusterBackups reference it.

However, looking into this led to me finding that we'll actually log an error here. Every time that a DBClusterBackup that references a given BackupStorage is deleted we'll log an erro if that BackupStorage is still in use, which isn't great.

recharte (approved):

api/database_cluster_backup.go
michal-kralik: Good catch, we did not handle errors properly. Fixed now.

recharte (approved):

PR #127, opened 2023-08-22
EVEREST-204 Update error message

oksana-grishchenko:
EVEREST-204 The old message was found confusing during the QA


michal-kralik (approved):
Would it make sense to update this during backup storage creation as well?

recharte (approved):

PR #128, opened 2023-08-23
EVEREST-250: redesign db cluster restore

michal-kralik:
EVEREST-250

- Requires changes EVEREST-249 to operator from https://github.com/percona/everest-operator/tree/EVEREST-249-restores-redesign


gen1us2k (approved):

api/database_cluster_restore.go
```diff
@@ -16,21 +16,100 @@
...
+	val := url.Values{}
+	val.Add("labelSelector", fmt.Sprintf("clusterName=%s", name))
+	req.URL.RawQuery = val.Encode()
+	path := req.URL.Path
+	// trim restores
+	path = strings.TrimSuffix(path, "/restores")
+	// trim name
+	path = strings.TrimSuffix(path, name)
+	path = strings.ReplaceAll(path, "database-clusters", "database-cluster-restores")
+	req.URL.Path = path
```
recharte, line 45: Do we need all this trimming? Wouldn't it be less error prone to build the URL path from scratch? I believe that the `kubernetesID` is all we need and we have it.

recharte (approved):
Well done @michal-kralik and @oksana-grishchenko
Left a minor suggestion, but I leave it up to you to decide whether or not to implement it, not a blocker.

api/database_cluster_restore.go
oksana-grishchenko: A good point, I'll address this in another PR along with the api-tests. I'm gonna merge this one to unblock the FE

PR #129, opened 2023-08-24
EVEREST-304: store namespace uid

michal-kralik:
EVEREST-304

- This can be used to identify the kubernetes cluster
- Requires https://github.com/percona/percona-everest-cli/pull/112 for tests to pass


api/kubernetes.go
```diff
@@ -47,6 +48,7 @@ func (e *EverestServer) ListKubernetesClusters(ctx echo.Context) error {
 			Id:        k.ID,
 			Name:      k.Name,
 			Namespace: k.Namespace,
+			Uid:       k.UID,
```
recharte, line 51: Should we name this `NamespaceUID` or something similar to make it clear that this ID is actually of the namespace and not the cluster?

michal-kralik: I suggest we keep it `UID` as there is no such thing as k8s cluster UID. We need to pick a resource and get its UID. The namespace is the most sensible thing I believe here.

oksana-grishchenko: Does it make sense to use that `UID` value as the `Id` and do not create an additional field? We generate the `Id` as `uuid.NewString()` anyway, so looks like there's not much value in keeping both

recharte: I guess we could. Even when we support deploying DBCs to other namespaces we'll still have the everest operator deployed in a single namespace so I guess we can use that UID as a k8s ID.

recharte: The only problem is that we can ensure that 2 different k8s won't, by chance, have the same UID.
Highly unlikely due to UUID format but we can't guarantee it won't happen.

michal-kralik: The cluster's UID is used in case we try to delete a cluster from Everest. At this point the user needs to provide a kubeconfig file. If the provided kubeconfig file is not related to the k8s cluster we have on file, we show a warning.

In order to be able to show the warning, we need to have a way to identify a k8s cluster. This field is for that.
Personally I would not mix the `ID` and `UID`  fields together since they both server a different purpose.

If there's a `UID` collision, which is highly unlikely, the worst case is we don't show a warning during cluster removal and remove the monitoring resources Everest installed from that cluster.

gen1us2k, line 51: I'm not sure that having the UID of a namespace will help us solve force removing k8s cluster from Everest at all.  CLI tries to remove monitoring resources assuming that the monitoring feature was enabled for this k8s cluster. Everest CLI or Everest backend does not have this knowledge, hence, during the unprovisioning it will fail to remove non-existent resources. 

Right now it's the only reason we need to have access to the k8s cluster if I'm not mistaken because CLI removes only monitoring operators/agents from the k8s cluster. I do not understand how namespace UID will help Everest to remove/do not remove VMagent resources from the k8s cluster. 

Since Everest needs to remove only monitoring stuff from the k8s cluster we need to identify monitoring enablement for this cluster and not run the cleanup process against the k8s cluster and just call the unregister endpoint that deletes the k8s cluster entries from internal storage.

michal-kralik: Per the description in the PR the UID is used only to identify the k8s cluster.
It is used to display a warning in case the provided kubeconfig does not match the selected cluster from Everest.

See the related PR for implementation details https://github.com/percona/percona-everest-cli/pull/111

gen1us2k: We can match the server address and namespace name. We have kubeconfig on the BE side stored in the secret storage and it contains a server field that includes the address. Server address is a better solution to match k8s clusters against each other and against the provided kubeconfig

michal-kralik: I believe UID approach is more consistent.
How would we approach a situation when the server address changes? E.g. due to a hostname change?

gen1us2k: Yeah. It may be consistent but I'd prefer to reuse existing entities to solve the issue. In the case of hostname change it's a matter of the different scope for having maintenance features.

gen1us2k: However, the UID approach does not guarantee duplicates across different k8s clusters that you discussed in the different thread. However, server address is a way more consistent and we can't have duplicates in server addresses

gen1us2k:
@oksana-grishchenko 

>the hostname approach - no new fields, no duplication, but need to handle the hostname changes.

It's not related to the hostname approach. We do not need to handle it right now and handling hostname changes it'll be requirements for it maybe later and it's out of scope right now

api/kubernetes.go
recharte: > How would we approach a situation when the server address changes? E.g. due to a hostname change?
If the server address changes, we lose access to the cluster. Currently, we don't have a way of changing the registered cluster so having an immutable server address is a must. I agree with Andrew that this is actually the best way of identifying if we're connecting to the correct cluster.

recharte (approved):

PR #137, opened 2023-08-31
EVEREST-330 Fix backup storage secret format

recharte:
EVEREST-330
We were creating the k8s secret with the everest internal secret IDs as the keys while it expects `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.


gen1us2k (approved):

oksana-grishchenko (approved):
Wow, I was surprised how come I never noticed that before. I tested storage locations and backups separately and for the backups I always applied an existing yaml with the secret CR🙈 Thanks for fixing this!

PR #147, opened 2023-09-06
EVEREST-384 [TECH DEBT] add validation unit-tests

oksana-grishchenko:
EVEREST-384 [TECH DEBT] add unit-tests

### DIFF
--- a/api/validation_test.go
+++ b/api/validation_test.go
@@ -78,3 +78,55 @@ func TestValidateRFC1123(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateCreateDatabaseClusterRequest(t *testing.T) {
+	t.Parallel()
+	type testCase struct {
+		name  string
+		value DatabaseCluster
+		err   error
+	}
+
+	cases := []testCase{
+		{
+			name:  "empty metadata",
+			value: DatabaseCluster{},
+			err:   errDBCEmptyMetadata,
+		},
+		{
+			name:  "no dbCluster name",
+			value: DatabaseCluster{Metadata: &map[string]interface{}{}},
+			err:   errDBCNameEmpty,
+		},
+		{
+			name: "empty dbCluster name",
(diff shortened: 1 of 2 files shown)


api/validation_test.go
```diff
@@ -78,3 +78,55 @@ func TestValidateRFC1123(t *testing.T) {
...
+			}},
+			err: errDBCNameWrongFormat,
+		},
+		{
+			name: "dbCluster name too long",
+			value: DatabaseCluster{Metadata: &map[string]interface{}{
+				"name": "a123456789a123456789a12",
+			}},
+			err: ErrNameTooLong("metadata.name"),
+		},
```
recharte, line 121: Can we also add tests that cover this code snippet?
```
	rfc1123Regex := "^[a-z0-9]([-a-z0-9]{0,61}[a-z0-9])?$"
	re := regexp.MustCompile(rfc1123Regex)
	if !re.MatchString(s) {
		return ErrNameNotRFC1123Compatible(name)
	}
```

Namely check:
* contain only lowercase alphanumeric characters or '-'
* start with an alphanumeric character
* end with an alphanumeric character

gen1us2k (approved):
Thanks. Looks good but take a look at Diogo's comment

api/validation_test.go
oksana-grishchenko: we have the tests for the `validateRFC1123` function separately [here](https://github.com/percona/percona-everest-backend/blob/5042b7b64a6a7b0a2aa0ad9b1cb2a86f35959f10/api/validation_test.go#L23-L80), it covers the different input for the regexp

oksana-grishchenko: however I added several new cases for the new function too to cover the code snippet, thanks

recharte: ohhh I forgot about those. Well in that case I wouldn't have pushed for this but it doesn't hurt either. Thanks for adding them regardless :hugs:

recharte (approved):

PR #158, opened 2023-09-14
EVEREST-435 Ensure Backend does not run uncontrollable goroutines

gen1us2k:
[EVEREST-435](https://jira.percona.com/browse/EVEREST-435) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-435

The backend runs a lot of goroutines without cancellation/synchronization and it's unclear when they exit

**Related pull requests**

- [link]

**Cause:**

**Solution:**
Since Everest supports only one kubernetes installation it does not make any sense to perform operations against kubernetes clusters asynchronously and it has a lot of side effects such as broken graceful shutdown which may lead to data inconsistency across k8s and storages. 

**CHECKLIST**
---
**Jira**

**Tests**
- [x] Are unit tests added where appropriate?

### DIFF
--- a/api/kubernetes.go
+++ b/api/kubernetes.go
@@ -71,7 +71,7 @@ func (e *EverestServer) RegisterKubernetesCluster(ctx echo.Context) error {
 	}
 	c := ctx.Request().Context()
 
-	_, err := clientcmd.BuildConfigFromKubeconfigGetter("", newConfigGetter(params.Kubeconfig).loadFromString)
+	_, err = clientcmd.BuildConfigFromKubeconfigGetter("", newConfigGetter(params.Kubeconfig).loadFromString)
 	if err != nil {
 		e.l.Error(err)
 		return ctx.JSON(http.StatusInternalServerError, Error{
--- a/api/kubernetes.go
+++ b/api/kubernetes.go
@@ -63,7 +63,7 @@ func (e *EverestServer) RegisterKubernetesCluster(ctx echo.Context) error {
 		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Could not list Kubernetes clusters")})
 	}
 	if len(list) != 0 {
-		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Everest does not support multi kubernetes clusters right now. Please delete existing clusters before registering a new one")})
+		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Everest does not support multiple kubernetes clusters right now. Please delete the existing cluster before registering a new one")})
 	}
 	var params CreateKubernetesClusterParams
 	if err := ctx.Bind(&params); err != nil {
--- a/api/kubernetes.go
+++ b/api/kubernetes.go
@@ -257,24 +257,23 @@ func (e *EverestServer) disableK8sClusterMonitoring(ctx echo.Context, kubeClient
 		})
 	}
 
-	go func() {
-		ctx := context.Background()
-		for _, s := range kubeClient.SecretNamesFromVMAgent(vmAgent) {
-			mcs, err := kubeClient.GetMonitoringConfigsBySecretName(ctx, s)
-			if err != nil {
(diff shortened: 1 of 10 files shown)


api/monitoring_instance.go
```diff
@@ -187,12 +187,12 @@ func (e *EverestServer) DeleteMonitoringInstance(ctx echo.Context, name string)
...
-		if err != nil {
-			e.l.Warn(errors.Wrapf(err, "could not delete monitoring instance API key secret %s", i.APIKeySecretID))
-		}
-	}()
+	_, err = e.secretsStorage.DeleteSecret(context.Background(), i.APIKeySecretID)
+	if err != nil {
+		err = errors.Wrapf(err, "could not delete monitoring instance API key secret %s", i.APIKeySecretID)
+		e.l.Warn(err)
+		return ctx.JSON(http.StatusInternalServerError, Error{Message: pointer.ToString(err.Error())})
+	}
```
recharte, line 195: I'm leaving this comment here but this is valid for most other go routines removed from in this PR.
Do we really want to return an InternalServerError when the removal of the monitoring instance succeeded and we failed to delete the secret?
Putting myself in the shoes of an API user, if I received a 500 when trying to delete a monitoring instance I would try again that request and I would find out that the monitoring instance was actually correctly deleted.
I'm not sure this is correct behavior...

gen1us2k: I fixed it and changed the logic for the following reasons
1. It's considered a bad practice to make network calls in a transaction
2. Deleting a secret should be in the transaction as well

api/monitoring_instance.go
```diff
@@ -163,22 +161,39 @@ func (e *EverestServer) DeleteMonitoringInstance(ctx echo.Context, name string)
...
+		return ctx.JSON(http.StatusInternalServerError, Error{Message: pointer.ToString("Could not make connection to the kubernetes cluster")})
+	}
+
+	err = kubeClient.DeleteConfig(ctx.Request().Context(), i, func(ctx context.Context, name string) (bool, error) {
+		return kubernetes.IsMonitoringConfigInUse(ctx, name, kubeClient)
+	})
+	if err != nil && !errors.Is(err, kubernetes.ErrConfigInUse) {
+		e.l.Error(errors.Wrap(err, "could not delete monitoring config from kubernetes cluster"))
+		return ctx.JSON(http.StatusInternalServerError, Error{Message: pointer.ToString("Could not delete monitoring config from the Kubernetes cluster")})
+	}
```
oksana-grishchenko, line 184: would it be an option here to run the code in a loop for all available clusters? It doesn't seem to be that much of a change but it would allow to work with multi k8s by default and reduce the risk to forget about this part in the future

oksana-grishchenko (approved):

api/kubernetes.go
```diff
@@ -57,13 +57,21 @@ func (e *EverestServer) ListKubernetesClusters(ctx echo.Context) error {
 
 // RegisterKubernetesCluster registers a k8s cluster in Everest server.
 func (e *EverestServer) RegisterKubernetesCluster(ctx echo.Context) error {
+	list, err := e.storage.ListKubernetesClusters(ctx.Request().Context())
+	if err != nil {
+		e.l.Error(err)
+		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Could not list Kubernetes clusters")})
+	}
+	if len(list) != 0 {
+		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Everest does not support multi kubernetes clusters right now. Please delete existing clusters before registering a new one")})
```
recharte, line 66:
```suggestion
		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Everest does not support multiple kubernetes clusters right now. Please delete the existing cluster before registering a new one")})
```

recharte (approved):

PR #160, opened 2023-09-14
EVEREST-428-rc-pipeline

oksana-grishchenko:
[EVEREST-428](https://jira.percona.com/browse/EVEREST-428) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**RC pipleline**
---
**Problem:**
EVEREST-428

Everest needs a workflow for building release branches.

**Solution:**
Create a manually triggering workflow. It would accept the version we want to create the RC for and would create the `release-*.*.*` branches in the backend, cli, frontend and catalog repos. It will update tags in the deployment scripts (so no manual intervention needed to run it from the rc-branch), build and push the new image to the registry.

### DIFF
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -14,8 +14,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
-      VERSION: ${{ github.event.inputs.version }}
-      RC_BRANCH: release-${{ github.event.inputs.version#v }}
+      #VERSION: ${{ github.event.inputs.version }}
+      VERSION: v100.100.100
+      RC_BRANCH: release-$VERSION#v
     steps:
       - name: Validate input
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/rc.yml
```diff
@@ -0,0 +1,20 @@
+name: Build and push RC image
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: An image tag
+        required: false
```
gen1us2k: should be true and match `v[0-9]+.[0-9]+.[0-9]+*`

.github/workflows/rc.yml
```diff
@@ -0,0 +1,185 @@
...
+          
+          if [[ -z ${check_branch} ]]; then
+            git checkout -b $RC_BRANCH
+            
+            # update tag refs in scripts
+            sed -i "s/dev-latest/$RC_BRANCH/g" deploy/quickstart-compose.yml deploy/quickstart-k8s.yaml
+            
+            # configure userdata for commits
+            git config --global user.email "oksana.grishchenko@percona.com"
+            git config --global user.name "Everest RC CI"
```
oksana-grishchenko, line 152: it would be great to have a technical email here, but I'm not sure which one to use. 
Instead of `Everest RC CI` we can be more specific and use `${{ github.actor }}` which returns the github username of the account that triggered the workflow. WDYT?

oksana-grishchenko:
1. Do we need to duplicate the required checks in the RC pipeline? I suggested to add this requirement
> Required checks should pass before triggering the job

but on the second thought it's gonna be hard to maintain the duplication and it doesn't bring that much value since we have that checks running on every push to `main` anyway 

2.
> The pipeline should be triggered on any new commit across backend and frontend repos to rebuild the release image

Once we agreed on this workflow, I'm gonna add to the FE repo the simplified version of this workflow which is gonna be triggered by any commit to the `release-x.x.x` branch and will rebuild the rc-image.

.github/workflows/rc.yml
gen1us2k: Yep. I thought about the same after adding this comment. Makes sense. Thanks

.github/workflows/rc.yml
gen1us2k: Makes sense. Let's try that

gen1us2k:
>Required checks should pass before triggering the job

We can skip this and be sure that pipelines are passing before triggering the job

gen1us2k (approved):
Approving this and keeping an eye on the IT ticket. Please be sure to merge it clean

PR #168, opened 2023-09-19
EVEREST-444 Add configurable bind address for the everest container

recharte:
[EVEREST-444](https://jira.percona.com/browse/EVEREST-444) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**Problem:**
In [EVEREST-333](https://jira.percona.com/browse/EVEREST-333) (#142) we identified and fixed a security vulnerability where we were exposing everest's port to 0.0.0.0 by default.
However, this fix broke the ability for users to access everest from outside localhost which is must have.

**Related pull requests**

- #142

**Cause:**
#142 

**Solution:**
Add a `EVEREST_BIND_ADDR` env var which is `127.0.0.1` by default but allows for users to control the bind address and thus the exposure of the everest container.

**CHECKLIST**
---
**Jira**

**Tests**
- ~[ ] Is an Integration test/test case added for the new feature/change?~
- ~[ ] Are unit tests added where appropriate?~

### DIFF
--- a/deploy/quickstart-compose.yml
+++ b/deploy/quickstart-compose.yml
@@ -18,4 +18,4 @@ services:
     environment:
       - DSN=postgres://admin:pwd@pg:5432/postgres?sslmode=disable
     ports:
-      - 8080:8080
+      - ${EVEREST_BIND_ADDR:-127.0.0.1}:8080:8080
--- a/deploy/quickstart-compose.yml
+++ b/deploy/quickstart-compose.yml
@@ -18,4 +18,4 @@ services:
     environment:
       - DSN=postgres://admin:pwd@pg:5432/postgres?sslmode=disable
     ports:
-      - 127.0.0.1:8080:8080
+      - 8080:8080


gen1us2k (commented):
What about having an env variable to set bind addr and by-default it will be 0.0.0.0?

recharte:
> What about having an env variable to set bind addr and by-default it will be 0.0.0.0?

I thought about that as the most likely solution but using 127.0.0.1 as default. But I was planning on doing that in the scope of [EVEREST-445](https://jira.percona.com/browse/EVEREST-445).

I separated this into two steps to make sure we can get this workaround right away and then tackle it properly in a follow-up PR because the full solution shall also be well documented in the installation instructions and I didn't know if we would be able to do it right away.

gen1us2k:
@recharte I added an env var. I'll test it soon

gen1us2k:
Tested and it works

deploy/quickstart-compose.yml
```diff
@@ -18,4 +18,4 @@ services:
     environment:
       - DSN=postgres://admin:pwd@pg:5432/postgres?sslmode=disable
     ports:
-      - 127.0.0.1:8080:8080
+      - ${EVEREST_BIND_ADDR:-127.0.0.1}:8080:8080
```
gen1us2k, line 21: Might not be the perfect one but it can be renamed

gen1us2k (approved):

recharte:
Thanks @gen1us2k, your commit looks good.
Since you went ahead and added the env var, I'll update the EVEREST-444 and EVEREST-445 descriptions accordingly and I'll poke the tech writing team to work on the docs for this then.

PR #169, opened 2023-09-19
EVEREST-302 Improve validation for backupStorage

gen1us2k:
[EVEREST-302](https://jira.percona.com/browse/EVEREST-302) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-302

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**


oksana-grishchenko:
That makes API a bit weird. Why do we have exactly that 3 parameters required? It doesn't really solve the original problem described in the ticket, I mean, what if I'll do a request with the "regioon" instead of "region"? I also have a concern about the FE compatibility. We need to make sure that it's not gonna break the existing integration before 0.3 or we need to create a FE ticket for this. 

Not sure we need to address the original problem though since we have the openAPI spec that describes the correct field names.

gen1us2k:
I'm not sure either but this PR was a low-effort stuff so I decided to discuss it with you. I'll talk to QA team

PR #198, opened 2023-09-26
EVEREST-457 Fixed backup deletion 

gen1us2k:
[EVEREST-457](https://jira.percona.com/browse/EVEREST-457) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-457 

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**


api/backup_storage.go
```diff
@@ -163,15 +163,15 @@ func (e *EverestServer) DeleteBackupStorage(ctx echo.Context, backupStorageName
...
 		e.l.Error(errors.Join(err, errors.New("could not init kube client for config")))
-		return nil
+		return ctx.JSON(http.StatusInternalServerError, Error{Message: pointer.ToString("Could not connect to the Kubernetes cluster")})
 	}
 
 	err = kubeClient.DeleteConfig(ctx.Request().Context(), bs, func(ctx context.Context, name string) (bool, error) {
 		return kubernetes.IsBackupStorageConfigInUse(ctx, name, kubeClient)
 	})
-	if err != nil && !errors.Is(err, kubernetes.ErrConfigInUse) {
+	if err != nil {
```
recharte, line 172: @gen1us2k I was thinking whether we should return something different from 500 if the error is `ErrConfigInUse`.
If the users try to delete a backup storage that is in use that's a client's error, it should be in the 4xx range not in the 5xx.
What are your thoughts?

gen1us2k: 4xx errors are for client errors and 5xx errors are for the server error. Here we have a server error hence the 501 error code. 
I thought about the same but I decided to go with the simplest option

gen1us2k: Anyway it will be better to return 400 in that case

recharte (approved):

PR #202, opened 2023-09-29
EVEREST-482 Removed nolint:funlen,cyclop directives

gen1us2k:
[EVEREST-482](https://jira.percona.com/browse/EVEREST-482) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-482

We shut some linters intentionally in the past 

**Cause:**

**Solution:**

The linter complains about cyclomatic complexity of 11 or 12 and I didn't see any other option instead of increasing this to 15. 15 is good enough and I had the setting for 30 in the past. Having the default setting for funlen and max-complexity set to 15 will prevent us from writing huge functions.

I refactored functions that had funlen issues but the rest of functions with `//nolint:cyclop` they had a cyclomatic complexity of 11 or 12 and there's no sane way to refactor them 

**CHECKLIST**
---
**Jira**

**Tests**


.golangci.yml
```diff
@@ -2,6 +2,9 @@
 # Almost all linters; some of them are optional.
 
 linters-settings:
+  cyclop:
+    max-complexity: 15
```
recharte, line 6: This seems like a cheat :see_no_evil:

recharte:
Can we start being strict with the PR description from now on? I'm seeing many PRs being opened with the default template.
I know that this particular PR doesn't add any new functionality or fix any bug but if we slack on these PRs we'll start slacking on any PR. Let's do a team effort to fill the template correctly for every PR.

gen1us2k:
>Can we start being strict with the PR description from now on? I'm seeing many PRs being opened with the default template.
> I know that this particular PR doesn't add any new functionality or fix any bug but if we slack on these PRs we'll start slacking on any PR. Let's do a team effort to fill the template correctly for every PR.

I agree. I was in the middle of filling it

.golangci.yml
recharte: Do you think that by quoting Linkin Park I'll become so numb that I can't feel you there? It seems like you're holding too tightly, afraid to lose control... :see_no_evil: :joy:

Jokes aside... Maybe we can actually split some of these further, I can try to take a look as well...

oksana-grishchenko (approved):
I thinks it's still ok to have 15 as max complexity. In case of API method handlers we need to do many different things in it and sometimes splitting the method into smaller brings more confusion than if we have all logic plain.

api/monitoring_instance.go
```diff
@@ -180,29 +180,36 @@ func (e *EverestServer) DeleteMonitoringInstance(ctx echo.Context, name string)
...
+	if err != nil {
+		return ctx.JSON(http.StatusInternalServerError, Error{
+			Message: pointer.ToString(err.Error()),
+		})
+	}
+
+	return ctx.NoContent(http.StatusNoContent)
+}
+
+func (e *EverestServer) deleteMonitoringConfig(c context.Context, i *model.MonitoringInstance) error {
```
michal-kralik, line 199:
```suggestion
func (e *EverestServer) deleteMonitoringConfig(ctx context.Context, i *model.MonitoringInstance) error {
```

michal-kralik (approved):

PR #207, opened 2023-10-02
EVEREST-478 Flaky tests

oksana-grishchenko:
[EVEREST-478](https://jira.percona.com/browse/EVEREST-478) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**Fix flaky tests**
---
**Problem:**
EVEREST-478 
In the backend repo there are several flaky tests for monitoring and backup-storages

**Solution**
Add random naming and cleanup


api-tests/tests/monitoring-instances.spec.ts
```diff
@@ -152,56 +142,45 @@ test('create monitoring instance missing pmm credentials', async ({ request }) =
...
-  expect(list.filter((i) => i.name.startsWith(`${namePrefix}${testPrefix}`)).length).toBe(3)
 
-  response = await request.delete(`/v1/monitoring-instances/${name}`)
-  expect(response.ok()).toBeTruthy()
-
-  response = await request.get('/v1/monitoring-instances')
-  expect(response.ok()).toBeTruthy()
-  list = await response.json()
-
-  expect(list.filter((i) => i.name.startsWith(`${namePrefix}${testPrefix}`)).length).toBe(2)
```
michal-kralik, line 199: This logic was in place intentionally.
I discovered we had a bug in the implementation where delete would delete the first instance, not the one selected by `name`.

I suggest we keep this in place.

oksana-grishchenko: Thanks for pointing that corner case out! The test is updated to handle it again. Also added a comment to avoid such confusions in the future.

api-tests/tests/monitoring-instances.spec.ts
```diff
@@ -13,365 +13,335 @@
...
 
-  expect(response.ok()).toBeTruthy()
-  const list = await response.json()
 
-  expect(list.filter((i) => i.name.startsWith(`${namePrefix}${testPrefix}`)).length).toBe(3)
+    await th.deleteMonitoringInstances(request, names)
 })
 
-test('get monitoring instance', async ({ request }) => {
-  const namePrefix = 'get-'
```
michal-kralik, line 168: This was in place for the same purpose - making sure we don't return the first entry instead of the correct one.

michal-kralik (approved):
I left a small comment about a potential issue with the tests.
Feel free to address if you think it's relevant.

oksana-grishchenko:
not relevant anymore, fixed by https://github.com/percona/percona-everest-backend/pull/431

PR #215, opened 2023-10-05
EVEREST-301 Shorten the error message for backup storage creation

gen1us2k:
[EVEREST-301](https://jira.percona.com/browse/EVEREST-301) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-301

On the UI error messages for backup storage cuts and the end user do not see that message

**Solution:**

Shorten error messages 

**CHECKLIST**
---
**Jira**

**Tests**


oksana-grishchenko (approved):

api/validation.go
```diff
@@ -266,12 +266,8 @@ func validateCreateBackupStorageRequest(ctx echo.Context, l *zap.SugaredLogger)
 
 	// check data access
 	if err := validateStorageAccessByCreate(params, l); err != nil {
-		if errors.Is(err, errUserFacingMsg) {
-			return nil, errors.Join(err, errors.New("could not connect to the backup storage, please check the new credentials are correct"))
-		}
-
 		l.Error(err)
-		return nil, errors.New("could not connect to the backup storage, please check the new credentials are correct")
+		return nil, err
```
michal-kralik, line 270: This will return cryptic error messages coming straight out of aws/azure/etc. sdk.
I think we need to provide a better experience.

gen1us2k: Nope. The message is not cryptic because the function returns user friendly message to the end user

michal-kralik: I see you fixed it now. Thank you

api/validation.go
```diff
@@ -266,12 +267,8 @@ func validateCreateBackupStorageRequest(ctx echo.Context, l *zap.SugaredLogger)
 
 	// check data access
 	if err := validateStorageAccessByCreate(params, l); err != nil {
-		if errors.Is(err, errUserFacingMsg) {
```
michal-kralik, line 269: By removing this line I don't think we need `errUserFacingMsg` anymore.

michal-kralik (approved):

PR #216, opened 2023-10-06
EVEREST-115: Azure storage support

michal-kralik:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-115

Added support for Azure storage.

**Related pull requests**

- https://github.com/percona/everest-operator/pull/99

**CHECKLIST**
---
**Jira**

**Tests**


api-tests/tests/backup-storages.spec.ts
```diff
@@ -92,9 +89,78 @@ test('add/list/get/delete backup storage success', async ({ request }) => {
...
+  }
+  const updated = await request.patch(`/v1/backup-storages/${name}`, {
+    data: updatePayload,
+  })
+
+  expect(updated.ok()).toBeTruthy()
+  const result = await updated.json()
+
+  expect(result.bucketName).toBe(updatePayload.bucketName)
+  expect(result.region).toBe(created.region)
```
oksana-grishchenko, line 146: we check the `region` that wasn't changed but do not check the `bucketName` that was changed

oksana-grishchenko (approved):
🚀

docs/spec/openapi.yml
```diff
@@ -1001,7 +1001,30 @@ paths:
...
-      description: Create a new backup storage object
+      description: |
+        Create a new backup storage object.
+
+        **Examples**:
+          ```
+          {
+            "name": "s3-storage",
+            "type": "s3",
+            "bucketName": "bucket1",
```
gen1us2k, line 1012: Spec is different between backend and operator

api-tests/tests/backup-storages.spec.ts
michal-kralik: We check the `bucketName` is updated on line 145.
We also check the `region` did not change during the update.

Am I missing anything?

docs/spec/openapi.yml
gen1us2k: https://github.com/percona/percona-everest-backend/blob/main/CONTRIBUTING.md#add-a-new-proxy-method Oksana described the way to add a new endpoint. You can use the same approach to update existing specs by copying oapispec from a CRD. I compared two specs in the operator and in the backend and they are different right now

michal-kralik: I'm really confused here. I don't understand which spec you're referring to.
Do you mean `DatabaseClusterBackup` or some other object?

The [operator PR](https://github.com/percona/everest-operator/pull/99) changed the `BackupStorage` spec which we don't import in the backend, hence no update is necessary. Is there any other spec you had in mind?

gen1us2k (approved):
Looks good. but please fix tests

PR #226, opened 2023-10-09
EVEREST-506 Dev config

oksana-grishchenko:
[EVEREST-506](https://jira.percona.com/browse/EVEREST-506) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**Dev config**
---
**Problem:**
EVEREST-506

Need to configure the telemetry related parameters for the development & testing purposes 

**Solution:**
Added the flags for the backend app. 
Didn't use the `go:build` tags for this purpose as we did for the `Debug` parameter, bc it doesn't scale well when we have many options to be configured, so moved the `Debug` parameter to the flags too. 

**CHECKLIST**
---
**Jira**

**Tests**


cmd/main.go
```diff
@@ -78,3 +81,10 @@ func main() {
 
 	l.Info("Exiting")
 }
+
+func initFlags() {
```
michal-kralik, line 85: Do we want to move over the other flags as well? E.g. `HTTP_PORT` etc?

oksana-grishchenko: I concidered using the existing config but thought it could be confusing from the ops standpoint when you have the dev options and the options that are needed for Everest to function properly are mixed together. But from second thought it's not so bad and the telemetry parameters are something that is needed for Everest to function, we just put an option to make it configurable. 

Regarding the Debug - I would leave it where it was then. It's not the property of the Everest server config, it's the property of the whole BE application and we need it in the different parts of the code independently of the Everest server. 

Thanks for bringing up the concern, moved the telemetry parameters to the Everest config.

cmd/config/config.go
```diff
@@ -31,3 +31,13 @@ func ParseConfig() (*EverestConfig, error) {
...
 }
+
+//nolint:gochecknoglobals
+var (
+	// Debug indicates if the current build is a debug build.
+	Debug bool
+	// TelemetryURL Everest telemetry endpoint.
+	TelemetryURL string
+	// TelemetryInterval Everest telemetry sending frequency.
+	TelemetryInterval string
```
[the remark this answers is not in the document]
oksana-grishchenko: Thanks for bringing up the concern, in [the comment](https://github.com/percona/percona-everest-backend/pull/226#discussion_r1351986471) are more details about the initial thoughts, but on the second thoughts I agree we should move the telemetry parameters to the Everest config.

recharte (approved):

michal-kralik (approved):

PR #229, opened 2023-10-11
EVEREST-506 Prod telemetry

oksana-grishchenko:
[EVEREST-506](https://jira.percona.com/browse/EVEREST-506) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**Distinguish release builds**
---
**Problem:**
EVEREST-506

With the current TelemetryURL configuration the test/dev data can still appear in prod telemetry (when testing dev-latest or rc builds). 

**Solution:**

Do not set the prod TelemetryURL as default. 
Set the url to the prod value _only_ when building a release image, so by default no url is set at all. 
For the testing purposes use the `make run-debug` which runs the version with the _dev_ telemetry enabled. 

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,13 @@ WORKDIR /everest
 
 COPY . .
 
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o /everest-api cmd/main.go
+ARG TELEMETRY_URL="https://check.percona.com"
+ARG TELEMETRY_INTERVAL="24h"
+
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -ldflags " \
+    -X 'github.com/percona/percona-everest-backend/cmd/config.TelemetryURL=$TELEMETRY_URL'  \
+    -X 'github.com/percona/percona-everest-backend/cmd/config.TelemetryInterval=$TELEMETRY_INTERVAL'" \
+    -o /everest-api cmd/main.go
 RUN apk add -U --no-cache ca-certificates
 
 FROM scratch
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,13 +4,22 @@ WORKDIR /everest
 
 COPY . .
 
+ARG IS_RELEASE
+
 ARG TELEMETRY_URL="https://check.percona.com"
 ARG TELEMETRY_INTERVAL="24h"
 
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -ldflags " \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -4,7 +4,22 @@ WORKDIR /everest
 
 COPY . .
 
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o /everest-api cmd/main.go
+ARG IS_RELEASE
```
gen1us2k, line 7: As the suggestion we can move the build to Makefile and copy the final binary into the container to make it simpler by calling 
```
make release
make debug
```
Yet it requires the changes to all build pipelines we have right now and this effort is out of scope for now. This solution works yet we need to create a ticket for our pipelines to fix it later and not keep it as is.

cmd/config/config.go
```diff
@@ -18,20 +18,37 @@ package config
...
 	// TelemetryInterval Everest telemetry sending frequency.
-	TelemetryInterval string `default:"24h" envconfig:"TELEMETRY_INTERVAL"`
+	TelemetryInterval string `envconfig:"TELEMETRY_INTERVAL"`
 }
 
 // ParseConfig parses env vars and fills EverestConfig.
 func ParseConfig() (*EverestConfig, error) {
 	c := &EverestConfig{}
 	err := envconfig.Process("", c)
+	if c.TelemetryURL == "" {
```
gen1us2k, line 46: The solution works fine but as we discussed we can have the opt-out of telemetry by using the following way

1. The end user needs to set TELEMETRY_URL to the empty string, like `export TELEMETRY_URL=`
2. The end user needs to restart a docker container

That's it. Telemetry has been disabled. But the requirement for that is that telemetry should not run if there's no URL specified to send metrics to. 

Right now it looks good but since the proposed way of opting out of telemetry needs to be discussed the changes look good for now but the proposed way of telemetry has the following benefits

- We can ship it faster because as the suggestion we had a proposal to have this in the CLI
- No effort is needed for the CLI and backend work to opt it out
- it guarantees that we won't mess the data from dev with prod because the prod URL sets only for release builds

gen1us2k (approved):
Great job. I left the explanation and additional context for the next reviewers but it looks great to me.

Dockerfile
oksana-grishchenko: Thanks for the suggestion, created a ticket for this improvement EVEREST-549

recharte (approved):

cmd/config/config.go
oksana-grishchenko: UPD: The opt-out via env var is discussed and confirmed within the team.

PR #235, opened 2023-10-13
EVEREST-549 Created make targets for the release

gen1us2k:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-549 

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,29 +4,13 @@ WORKDIR /everest
 
 COPY . .
 
-ARG IS_RELEASE
-
-ARG TELEMETRY_URL="https://check.percona.com"
-ARG TELEMETRY_INTERVAL="24h"
-
-RUN if [[ $IS_RELEASE = 1 ]]; then  \
-    # for the release builds set up the production telemetry parameters
-    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -ldflags "\
-            -X 'github.com/percona/percona-everest-backend/cmd/config.TelemetryURL=$TELEMETRY_URL'  \
-            -X 'github.com/percona/percona-everest-backend/cmd/config.TelemetryInterval=$TELEMETRY_INTERVAL'"  \
-        -o /everest-api cmd/main.go; \
-    else \
-     # for all the other builds no telemetry parameters are provided by default
-    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o /everest-api cmd/main.go; \
-fi
-
 RUN apk add -U --no-cache ca-certificates
 
 FROM scratch
 
 WORKDIR /
 
-COPY --from=build /everest-api /everest-api
+COPY ./bin/percona-everest-backend  /everest-api
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY migrations /migrations
 
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,13 @@
 FILES = $(shell find . -type f -name '*.go')
 
+RELEASE_VERSION ?= $(shell git describe --always --tags --dirty | cut -b2-)
+RELEASE_FULLCOMMIT ?= $(shell git rev-parse HEAD)
+FLAGS = -X 'github.com/percona/percona-everest-backend/pkg/version.ProjectName=everestctl' \
+	-X 'github.com/percona/percona-everest-backend/pkg/version.Version=$(RELEASE_VERSION)' \
+	-X 'github.com/percona/percona-everest-backend/pkg/version.FullCommit=$(RELEASE_FULLCOMMIT)' \
+
+LD_FLAGS = -ldflags " \ $(FLAGS) "
+
 default: help
 
 help:                   ## Display this help message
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -4,29 +4,13 @@ WORKDIR /everest
 
 COPY . .
 
-ARG IS_RELEASE
```
oksana-grishchenko, line 7: we can remove the build-args from the release.yml `Everest - Build and Push everest release image` step

Makefile
```diff
@@ -11,10 +19,15 @@ init:                   ## Install development tools
 	cd tools && go generate -x -tags=tools
 
 build:                ## Build binaries
-	go build -race -o bin/percona-everest-backend ./cmd
+	go build $(LD_FLAGS) -o bin/percona-everest-backend ./cmd
+
+release: FLAGS += -X 'github.com/percona/percona-everest-backend/cmd/config.TelemetryURL=https://check.percona.com/'\ -X 'github.com/percona/percona-everest-backend/cmd/config.TelemetryInterval=24h'
+
+release: build
```
oksana-grishchenko, line 26: that two targets with the same name look odd, didn't know it's possible to configure targets like this. But it works, cool

Makefile
```diff
@@ -1,5 +1,13 @@
 FILES = $(shell find . -type f -name '*.go')
 
+RELEASE_VERSION ?= $(shell git describe --always --tags --dirty | cut -b2-)
+RELEASE_FULLCOMMIT ?= $(shell git rev-parse HEAD)
+FLAGS = -X 'github.com/percona/percona-everest-backend/pkg/version.ProjectName=everestctl' \
```
oksana-grishchenko, line 5: A bit confused about the ProjectName. We referenced the cli as `everestctl`, now it's a project name for the BE app too...

gen1us2k: Copy pasted code

Makefile
gen1us2k: make file is just a legacy from the 70s and sometimes it the code looks weird

Makefile
gen1us2k: Fixed

oksana-grishchenko (approved):

recharte:
No description in the PR :cry:

gen1us2k:
@recharte added the description

Dockerfile
```diff
@@ -4,29 +4,13 @@ WORKDIR /everest
...
-    # for the release builds set up the production telemetry parameters
-    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -ldflags "\
-            -X 'github.com/percona/percona-everest-backend/cmd/config.TelemetryURL=$TELEMETRY_URL'  \
-            -X 'github.com/percona/percona-everest-backend/cmd/config.TelemetryInterval=$TELEMETRY_INTERVAL'"  \
-        -o /everest-api cmd/main.go; \
-    else \
-     # for all the other builds no telemetry parameters are provided by default
-    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o /everest-api cmd/main.go; \
-fi
-
```
recharte, line 22: What happened to the build? Why did you remove it entirely?

Dockerfile
```diff
@@ -4,29 +4,13 @@ WORKDIR /everest
...
 RUN apk add -U --no-cache ca-certificates
 
 FROM scratch
 
 WORKDIR /
 
-COPY --from=build /everest-api /everest-api
+COPY ./bin/percona-everest-backend  /everest-api
```
recharte, line 13: Looks to me like we aren't building this binary anywhere in this Dockerfile. Was this the intention? How can we be subject to whatever binary is stored in the bin directory?
I understand that `make build` generates this binary but I don't think that in order to build the container image this should be a pre-requisite. This will lead to errors when someone thinks it's building the newest version of the container where instead it's just using the old binary it had stored in the bin directory from months ago...

Dockerfile
gen1us2k: I moved to the 
```
make build
or
make release
```
on the host machine and then copy binary to the container

Dockerfile
gen1us2k: >This will lead to errors when someone thinks it's building the newest version of the container where instead it's just using the old binary it had stored in the bin directory from months ago...

No one builds the container manually and I added required changes to our pipelines

recharte: Is there any technical reason for having it in the makefile target and not on the dockerfile build stage?

michal-kralik (approved):

Dockerfile
gen1us2k: We had a discussion with @oksana-grishchenko about that [here](https://github.com/percona/percona-everest-backend/pull/229/files#r1355362669). Env vars looks weird in the pipelines and since we need to have different build tags as the first option to keep everything clean without env vars/branches the best option we found is we use
```
make build 
make run-debug # for dev cases
```
for dev builds and 
```
make release
```
for releasing artifacts

gen1us2k: And since we use Dockerfile in our pipelines mostly we can afford these changes because the user for Dockerfile is machine/CI/Build script

PR #245, opened 2023-10-16
EVEREST-400 encrypt secrets in secrets storage

recharte:
[EVEREST-400](https://jira.percona.com/browse/EVEREST-400) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**EVEREST-400 encrypt secrets in secrets storage**
---
**Problem:**
EVEREST-400

We were previously storing secrets in plain-text which obviously a really bad idea.

**Solution:**
Leverage hashicorp vault's AESGCM barrier module to encrypt the secrets in the postgres internal database.

**CHECKLIST**
---
**Jira**

**Tests**
- ~[] Is an Integration test/test case added for the new feature/change?~
- ~[] Are unit tests added where appropriate?~


cmd/config/config.go
```diff
@@ -57,5 +66,10 @@ func ParseConfig() (*EverestConfig, error) {
 		c.TelemetryInterval = TelemetryInterval
 	}
 
-	return c, err
+	rootKey, err := base64.StdEncoding.DecodeString(c.RootKey)
+	if err != nil || len(rootKey) != 2*aes.BlockSize {
```
michal-kralik, line 70: `2*aes.BlockSize` sounds like a magic number :)
Is this correct or just a coincidence? E.g. is there a more appropriate constant to be used?

deploy/quickstart-k8s.yaml
```diff
@@ -1,3 +1,10 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: everest-root-key
+data:
+  root-key: bXRlRWRuODNLeHFzM0xuVW9tcUZJelRRcEtwK256bm5aSytZSHlCVXRSWT0=
```
michal-kralik, line 6: Would it make sense to consider not having a constant here?
I expect most of our users not to change this manually and we end up with a "default" password situation.

model/secrets_storage.go
```diff
@@ -0,0 +1,94 @@
...
+	barrier *vault.AESGCMBarrier
+}
+
+// NewSecretsStorage returns a new SecretsStorage instance.
+func NewSecretsStorage(ctx context.Context, dsn string, key []byte) (*SecretsStorage, error) {
+	physical, err := physPostgreSQL.NewPostgreSQLBackend(
+		map[string]string{
+			"connection_url": dsn,
+			"table":          "secrets",
+			"ha_enabled":     "false",
```
michal-kralik, line 40: How would this config affect production HA deployments?

model/secrets_storage.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return "", err
+	}
+	return string(storedEntry.Value), nil
+}
+
+// UpdateSecret updates the secret by its id.
+func (secStor *SecretsStorage) UpdateSecret(ctx context.Context, id, value string) error {
+	return secStor.CreateSecret(ctx, id, value)
+}
```
michal-kralik, line 89: Since update is essentially "create", can we change our API with this PR?
I would suggest we have:
- `PutSecret`
- `GetSecret`
- `DeleteSecret`

I experienced create vs. update issues during user authentication development and had to resort to adding `PutSecret` as a helper.

I think this would better copy how secrets are stored in Vault.

api/everest.go
```diff
@@ -64,20 +87,51 @@ func NewEverestServer(c *config.EverestConfig, l *zap.SugaredLogger) (*EverestSe
...
+	e.secretsStorage, err = model.NewSecretsStorage(ctx, e.config.DSN, rootKey)
+	if err != nil {
+		return err
+	}
+
+	// DB schema version 8 is a transitional version that is used to
+	// automatically migrate the data from the old schema (v7) to the new one
+	// (v9). If the DB schema version is 8, we need to perform the secret data
+	// migration from the old schema to the new one. After that, we can apply
+	// the rest of the migrations.
```
michal-kralik, line 121: Just a note here - I understand this logic. I'm worried though the approach affects our ability to deploy Everest in HA due to race-conditions.

Once we deploy multiple replicas, we may run into issues during bootstrapping.
We can consider dropping this code after a couple versions when we believe everyone has migrated properly.

Alternatively this can be considered a breaking change and we can say we don't support upgrades at this point.

What do you think?

michal-kralik (commented):
Generally looks good. Left a couple comments for discussion.

api/everest.go
recharte: This is a great point that I definitely didn't consider. Let me expand this question though, wouldn't we have this issue already with the standard migrations? Are they safe to be done in a multi-replica scenario?

cmd/config/config.go
```diff
@@ -41,12 +44,18 @@ type EverestConfig struct {
 	TelemetryURL string `envconfig:"TELEMETRY_URL"`
 	// TelemetryInterval Everest telemetry sending frequency.
 	TelemetryInterval string `envconfig:"TELEMETRY_INTERVAL"`
+	// RootKey is a base64-encoded 256-bit key used for the secrets encryption.
+	RootKey string `required:"true" envconfig:"ROOT_KEY"`
```
michal-kralik, line 48: What do you think about being more specific with naming, e.g. `SECRETS_ROOT_KEY`?
Once we have a couple of such variables, e.g. a root key for auth, it would be great to be able to distinguish at first sight which root key we're referring to.

model/secrets_storage.go
recharte: I wasn't sure about the impact this config might have so I did a little digging.
Turns out that having ha_enabled set to true doesn't impact the methods used by the barrier API at all.
It provide an extra [HABackend interface](https://github.com/hashicorp/vault/blob/c2f5ee0c6251e519c14de5d89a4ce337c688cb10/sdk/physical/physical.go#L51-L61) where there's a `LockWith()` method that provides the users with a [lock](https://github.com/hashicorp/vault/blob/c2f5ee0c6251e519c14de5d89a4ce337c688cb10/sdk/physical/physical.go#L142-L154) that can then be manually locked or unlocked.
In case you're curious, you can have a look at how hashicorp uses the lock returned by `LockWith()` to protect a long running routine during the leader election process [here](https://github.com/hashicorp/vault/blob/c2f5ee0c6251e519c14de5d89a4ce337c688cb10/vault/ha.go#L470-L741)

I don't think we have a use case for this type of locking mechanism right now, so I'd keep it as false.

deploy/quickstart-k8s.yaml
gen1us2k, line 6: IMO we should highlight this in the documentation instead of providing something with hardcoded value. openssl rand might be helpful

model/secrets_storage.go
recharte: I tried to keep the same API we had but I completely agree that it makes more sense to change. I'll update it.

go.mod
```diff
@@ -27,68 +29,255 @@ require (
...
+	github.com/Azure/go-autorest/autorest/azure/cli v0.4.6 // indirect
+	github.com/Azure/go-autorest/autorest/date v0.3.0 // indirect
+	github.com/Azure/go-autorest/autorest/to v0.4.0 // indirect
+	github.com/Azure/go-autorest/autorest/validation v0.3.1 // indirect
+	github.com/Azure/go-autorest/logger v0.2.1 // indirect
+	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
+	github.com/BurntSushi/toml v1.3.2 // indirect
+	github.com/DataDog/datadog-go v3.2.0+incompatible // indirect
+	github.com/Jeffail/gabs v1.1.1 // indirect
+	github.com/Masterminds/goutils v1.1.1 // indirect
```
gen1us2k, line 54: I'm not completely sure that we can use Vault as a dependency in terms of recent news and changed license.  I assume we should talk to legal or IT about this. 

>The products using the BSL license from here forward are HashiCorp Terraform, Packer, Vault, Boundary, Consul, Nomad, Waypoint, and Vagrant. The latest releases under MPL are Terraform 1.5.5, Packer 1.9.2, Vault 1.14.1, Boundary 0.13.1, Consul 1.16.1, Nomad 1.6.1, Waypoint 0.11.4, and Vagrant 2.3.7, which is MIT licensed. Security fixes will be backported under MPL 2.0 through December 31, 2023.

Do we have any alternatives or forks?

gen1us2k: Did we have a discussion about this?

model/secrets_storage.go
gen1us2k: Agree with the suggestion

go.mod
recharte: Well Maxim did raise this question to Vadim on the last "Coffee and conversation with CTO" and from what he explain I interpreted as being completely fine because we do not fall into the category of a competing product.

.github/workflows/ci.yml
```diff
@@ -243,7 +249,7 @@ jobs:
         run: |
           make local-env-up
           make build-debug
-          make run-debug &> everest-backend.log &
+          ROOT_KEY="ke6IcFSyulOODrq4pK2ml9Jwu7y6MY8CM0PbT19Wv6M=" make run-debug &> everest-backend.log &
```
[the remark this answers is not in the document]
recharte: I wasn't sure about how ephemeral this deployment was so I hardcoded it, but I think you're right, generating it is probably better.

michal-kralik:
@recharte One other thing I just realized?
I was expecting we would connect to an external Vault instance - e.g. in case customers already run their own instance.

Do I understand it correctly that instead of running a separate instance we're running Vault as part of Everest backend?

recharte:
> @recharte One other thing I just realized. I was expecting we would connect to an external Vault instance - e.g. in case customers already run their own instance.
> 
> Do I understand it correctly that instead of running a separate instance we're running Vault as part of Everest backend?

Neither of those is what we want right now.

What you're describing is [phase 2](https://www.notion.so/percona/Percona-Secrets-Management-628bcc4415934953a4def3916e74a947?pvs=4#8d9548dda1d748e5ab31f352aca5f072) of the Percona secrets management plan. In phase 2, the plan is to connect to a vault instance which can be external and managed by the user or bundled with a percona product.

What we are implementing right now is [phase 1](https://www.notion.so/percona/Percona-Secrets-Management-628bcc4415934953a4def3916e74a947?pvs=4#cb3fc788e74b4c2883f43c6b58b93e12). This leverages the already existing backend DB (postgresql) and encrypts the secrets before persisting them in the DB.
We had many options for handling this encryption before storing but we decided on using the module used by vault to abstract this operation in simple Put, Get, Delete methods. So we are not using vault per se, we are using the same mechanism that vault uses to encrypt secret when using the postgresql physical backend.

cmd/config/config.go
recharte: I agree

deploy/quickstart-k8s.yaml
recharte: Do you think that having a placeholder value here like
```
data:
  root-key: <SECRETS_ROOT_KEY>
```

And then in the documentation ask users to do a sed replacement of this placeholder with the root key generated by them would be the best approach?

something like
```
export SECRETS_ROOT_KEY=$(openssl rand -base64 32)
echo "$SECRETS_ROOT_KEY"
sed -i "s/<SECRETS_ROOT_KEY>/$SECRETS_ROOT_KEY/" quickstart-k8s.yml
```

cc. @michal-kralik you commented the same, let's followup here

recharte: Let's continue this conversation [here](https://github.com/percona/percona-everest-backend/pull/245#discussion_r1362051108)

michal-kralik: I'm not too familiar with `kubectl` but it appears we cannot pass anything to `kubectl apply -f`.
If that's the case, maybe we could leave this empty - hence Everest will fail to start. There should be a nice error message.

Instead of instructing users to use the `quickstart.yml` directly, populating the `root-key` could be part of the `install.sh` or `helm` chart so we take care of it.
Or if this does not make sense, we would have to come up with some other approach.

cmd/config/config.go
gen1us2k: I have the same feeling and this is unclear to me why it should be not equal. Maybe add some explanation for it?

go.mod
gen1us2k: Yet we have guidelines with allowed-to-use licenses. I assume that we add 1.15 and it's still MPL but we should pin this version or do something else about this. I'm not sure what exactly.

model/database.go
```diff
@@ -89,21 +89,55 @@ func (db *Database) Transaction(fn func(tx *gorm.DB) error) error {
...
 		return 0, errors.Join(err, errors.New("failed to setup migrator"))
 	}
 
+	v, dirty, err := m.Version()
+	if err != nil && !errors.Is(err, migrate.ErrNilVersion) {
+		return 0, errors.Join(err, errors.New("failed to check version"))
+	}
+	if dirty {
+		return 0, errors.New("database is dirty; manual fix is required")
+	}
```
gen1us2k, line 108: Just out of my mind. But what about moving this into a single binary or a subcommand that the end user will run only once during the upgrade to 0.4? We will have documentation about upgrading to 0.4 and we can be sure that this migration run only once. WDYT?

api/everest.go
```diff
@@ -64,20 +87,51 @@ func NewEverestServer(c *config.EverestConfig, l *zap.SugaredLogger) (*EverestSe
...
+	if dbVersion == 8 {
```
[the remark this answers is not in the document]
gen1us2k: Okay. I understood this magic. Thanks. I left a comment below 

>Just out of my mind. But what about moving this into a single binary or a subcommand that the end user will run only once during the upgrade to 0.4? We will have documentation about upgrading to 0.4 and we can be sure that this migration run only once. WDYT?

deploy/quickstart-k8s.yaml
recharte: So are you suggesting that we should the secret entirely from the quickstart manifest and just add instructions on how to create one before applying the manifest? Also generate it using the install script.

gen1us2k: Yep. Until we have helm charts ready

model/database.go
recharte: So are you suggesting that we migrate the schema to v8 which doesn't drop the plain text secrets table; then the user can manually run a script that migrates the data; then somewhere in a future release we will add the v9 schema migration to drop that table?

cmd/config/config.go
recharte: Please check if 3a6aeb056bf50f1f83b52cf70089ce91664b3130 improved this

deploy/quickstart-k8s.yaml
recharte: Ok sounds good, I'll remove it

cmd/config/config.go
michal-kralik: eh, not what I had in mind :)
Wouldn't `aes` have a proper constant?

I mean it's not a big deal, it's just unclear why we want to multiply a regular block size by 2.

go.mod
recharte: Version 1.15.0 is already BSL but my understanding is that it's fine for us to use it.

model/database.go
gen1us2k: Not exactly. I'm talking about moving this logic to a single binary or `./bin/everest-backend migrate` subcommand and keeping it for a couple of releases until 0.7 with highlighting in the documentation and deprecating this after 0.4.

In that case, we can have Migrate logic in the everest backend untouched and we can add this binary to our docker container so it runs first later we will drop this since this is a one-time operation that affects users of 0.3 release and the users of 0.4 will get this feature without any migration needed from their side.

api/everest.go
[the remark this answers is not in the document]
gen1us2k: Let's close this one and continue below

model/database.go
recharte: I initially agreed with your suggested approach but I now found a drawback that I think undermines the benefits of the segregation onto a separate binary.

Adding this 2nd binary means that we need to change the entrypoint of the container to call a script (e.g. `entrypoint.sh`) that executes both commands. At first this seemed fine, however we're using a `scratch` container image which doesn't have a shell. Therefore, to make this work we must either change the base image or have the main binary execute a second binary. I think that I much rather have the current implementation than one of these 2 options.

Moreover, as discussed we can revert c09ba522061d3b6902db3e73762309bf5f92c279 as soon as 0.4.0 is released and just mention in the release notes of the subsequent couple of release that 0.4 introduced a breaking change and if users want to migrate from 0.3 to 0.5 or 0.6 they must first upgrade to 0.4.

Given the aforementioned pros and cons, I don't think that it's worth the effort to move this logic to a separate binary. Do you agree @gen1us2k?

api/everest.go
recharte: @michal-kralik As @gen1us2k said
> Everest does not support HA because of race conditions anyway and keeps the state for some operations in memory.

Also we are planning to remove this logic in the next release ([reference](https://github.com/percona/percona-everest-backend/pull/245#discussion_r1362561048)):
> we can revert https://github.com/percona/percona-everest-backend/commit/c09ba522061d3b6902db3e73762309bf5f92c279 as soon as 0.4.0 is released and just mention in the release notes of the subsequent couple of release that 0.4 introduced a breaking change and if users want to migrate from 0.3 to 0.5 or 0.6 they must first upgrade to 0.4.

deploy/quickstart-k8s.yaml
recharte: ac789c24dfa72ed82b6ae2acd73ba9450d3e92b7

.github/workflows/ci.yml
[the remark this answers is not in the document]
recharte: b020b2fcd5f183b309226d17ea6c3d1c751ba191

cmd/config/config.go
recharte: 24c85c7de263ed7d24617c12c9c34a9db9dd5d9e

model/secrets_storage.go
recharte: e3988b5cb0e17f776d5bbf3ad55dcc885ec5e0cf

cmd/config/config.go
```diff
@@ -44,8 +48,10 @@ type EverestConfig struct {
 	TelemetryURL string `envconfig:"TELEMETRY_URL"`
 	// TelemetryInterval Everest telemetry sending frequency.
 	TelemetryInterval string `envconfig:"TELEMETRY_INTERVAL"`
-	// RootKey is a base64-encoded 256-bit key used for the secrets encryption.
-	RootKey string `required:"true" envconfig:"ROOT_KEY"`
+	// DisableTelemetry disable Everest and the upstream operators telemetry
+	DisableTelemetry bool `default:"false" envconfig:"DISABLE_TELEMETRY"`
```
michal-kralik, line 52: I think I remember we intentionally wanted to leave this out. But I'm not sure.
@oksana-grishchenko any thoughts?

michal-kralik (approved):
lgtm

cmd/config/config.go
recharte: @michal-kralik this was just a merge with main, that code was added in #243

michal-kralik: awesome, I missed this.

oksana-grishchenko (approved):
Great job! Pretty tricky way to migrate the secrets storage, thanks for adding the comments. It should work though, haven't found any misalignment. Concerns that appeared during the review are already discussed and resolved in the existing threads, nothing to add here. 

If we plan to drop some of the migration logic in the later releases, I would suggest to create an Admin&Maintenance ticket to not forget about it.

model/database.go
gen1us2k: OK

recharte:
Hey folks, I'm keeping this PR open until Thursday, as I'm waiting for a confirmation from legal regarding the BSL license of vault@v1.15.0.

deploy/quickstart-compose.yml
```diff
@@ -17,5 +17,6 @@ services:
         condition: service_healthy
     environment:
       - DSN=postgres://admin:pwd@pg:5432/postgres?sslmode=disable
+      - ROOT_KEY=${ROOT_KEY}
```
maxkondr, line 20:
```suggestion
      - SECRETS_ROOT_KEY=${ROOT_KEY}
```

deploy/quickstart-k8s.yaml
```diff
@@ -29,6 +29,11 @@ spec:
           env:
             - name: DSN
               value: postgres://admin:pwd@localhost:5432/postgres?sslmode=disable
+            - name: ROOT_KEY
```
maxkondr, line 32:
```suggestion
            - name: SECRETS_ROOT_KEY
```

README.md
```diff
@@ -20,9 +20,17 @@ The Percona Everest has two primary components that assist you in creating the e
 
 To start using Everest, use the following commands:
 
+```sh
+export ROOT_KEY=$(openssl rand -base64 32)
+echo "$ROOT_KEY"
```
maxkondr, line 25:
```suggestion
export SECRETS_ROOT_KEY=$(openssl rand -base64 32)
echo "$SECRETS_ROOT_KEY"
```

api/everest.go
```diff
@@ -64,20 +87,51 @@ func NewEverestServer(c *config.EverestConfig, l *zap.SugaredLogger) (*EverestSe
 	if err := e.initHTTPServer(); err != nil {
 		return e, err
 	}
-	err := e.initEverest()
+	err := e.initEverest(context.Background())
```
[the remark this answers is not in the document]
recharte: That context is passed along to every method of the AES GCM barrier module. It can be used for many things as for example propagating cancellation signals.

deploy/quickstart-k8s.yaml
```diff
@@ -29,6 +29,11 @@ spec:
           env:
             - name: DSN
               value: postgres://admin:pwd@localhost:5432/postgres?sslmode=disable
+            - name: SECRETS_ROOT_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: everest-root-key
+                  key: root-key
```
maxkondr, line 36:
```suggestion
                  key: secrets-root-key
```

README.md
```diff
@@ -20,9 +20,17 @@ The Percona Everest has two primary components that assist you in creating the e
 
 To start using Everest, use the following commands:
 
+```sh
+export SECRETS_ROOT_KEY=$(openssl rand -base64 32)
+echo "$SECRETS_ROOT_KEY"
+```
+This generates a base64-encoded 256-bit key used for secrets encryption. Make
```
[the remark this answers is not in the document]
recharte: Well we need to have a string representation of a 32 bytes array, the most popular options are base64 and hex. Base64 seems fine and is also what's used by vault externally.

cmd/config/config.go
```diff
@@ -41,18 +50,30 @@ type EverestConfig struct {
...
 	}
 	if c.TelemetryInterval == "" {
 		c.TelemetryInterval = TelemetryInterval
 	}
 
-	return c, err
+	// SecretsRootKey must be a base64-encoded 256-bit key.
+	secretsRootKey, err := base64.StdEncoding.DecodeString(c.SecretsRootKey)
+	if err != nil || len(secretsRootKey) != AES256BitKeySize {
+		return nil, errors.New("secrets root key must be a base64-encoded 256-bit key")
```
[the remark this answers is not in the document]
recharte: Do you really feel it's worth separating this into to two error logs?

api/everest.go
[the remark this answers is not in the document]
recharte: Not necessarily, we just pass the context in the API in case any operation needs to be canceled. Most likely you won't need to cancel any on-going operation when shutting down.

api/kubernetes.go
```diff
@@ -106,7 +106,7 @@ func (e *EverestServer) RegisterKubernetesCluster(ctx echo.Context) error {
 		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Could not create Kubernetes cluster")})
 	}
 
-	err = e.secretsStorage.CreateSecret(c, k.ID, params.Kubeconfig)
+	err = e.secretsStorage.PutSecret(c, k.ID, params.Kubeconfig)
 	if err != nil {
 		e.l.Error(err)
 		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Could not store kubeconfig in secrets storage")})
```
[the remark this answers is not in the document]
recharte: You're right and like this one there's also other examples spread throughout the code. This is something we said we would revisit during the multi-k8s support.

gen1us2k (approved):

migrations/000009_secret_storage.down.sql
```diff
@@ -0,0 +1,9 @@
+DROP TABLE secrets;
```
michal-kralik, line 1: Do migrations work properly given there's no `00008` file?

recharte: Actually there is :slightly_smiling_face: see #252

michal-kralik: Great, I did not have my local repo updated.

PR #252, opened 2023-10-19
EVEREST-507 Report metrics

oksana-grishchenko:
[EVEREST-507](https://jira.percona.com/browse/EVEREST-507) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**Collect and report metrics**
---
**Problem:**
EVEREST-507

Report the db engines metrics - how many db clusters of each type are running. 

A step forward multi k8s support - generate own Everest ID instead of using the k8s cluster ID. 

The telemetry request [example](https://www.notion.so/percona/Pillars-Telemetry-Design-efea81c1f170456099503bbcdbd0ad9b?pvs=4#fb2130ee3a30425c9e3dd45045f34235) 

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/api/everest.go
+++ b/api/everest.go
@@ -28,6 +28,7 @@ import (
 	"sync"
 
 	"github.com/deepmap/oapi-codegen/pkg/middleware"
+	"github.com/google/uuid"
 	"github.com/labstack/echo/v4"
 	echomiddleware "github.com/labstack/echo/v4/middleware"
 	"go.uber.org/zap"
@@ -65,6 +66,31 @@ func NewEverestServer(c *config.EverestConfig, l *zap.SugaredLogger) (*EverestSe
 		return e, err
 	}
 	err := e.initEverest()
+	if err != nil {
+		e.l.Error(err)
+		return e, err
+	}
+
+	ctx := context.Background()
+
+	_, err = e.storage.CreateSettings(ctx, model.SettingsParams{ID: uuid.NewString()})
+	if err != nil {
+		e.l.Error(err)
+		return e, err
+	}
+
+	if !c.DisableTelemetry {
+		// To prevent leaking test data to prod,
+		// the prod TelemetryURL is set for the release builds during the build time.
+		// The dev TelemetryURL is set only when running `make run-debug`.
+		if c.TelemetryURL != "" {
+			go func() {
+				e.runTelemetryJob(ctx, c)
+			}()
+		} else {
+			e.l.Info("Telemetry is not running, the TELEMETRY_URL is not set")
+		}
+	}
 
 	return e, err
 }
--- a/api/everest.go
+++ b/api/everest.go
@@ -28,7 +28,6 @@ import (
 	"sync"
 
 	"github.com/deepmap/oapi-codegen/pkg/middleware"
-	"github.com/google/uuid"
 	"github.com/labstack/echo/v4"
 	echomiddleware "github.com/labstack/echo/v4/middleware"
 	"go.uber.org/zap"
@@ -73,7 +72,7 @@ func NewEverestServer(c *config.EverestConfig, l *zap.SugaredLogger) (*EverestSe
 
 	ctx := context.Background()
 
-	_, err = e.storage.CreateSettings(ctx, model.SettingsParams{ID: uuid.NewString()})
+	err = e.storage.InitSettings(ctx)
 	if err != nil {
 		e.l.Error(err)
 		return e, err
--- a/api/everest.go
+++ b/api/everest.go
@@ -70,25 +70,10 @@ func NewEverestServer(c *config.EverestConfig, l *zap.SugaredLogger) (*EverestSe
 		return e, err
 	}
 
-	ctx := context.Background()
-
-	err = e.storage.InitSettings(ctx)
+	err = e.storage.InitSettings(context.Background())
 	if err != nil {
 		e.l.Error(err)
 		return e, err
-	}
-
-	if !c.DisableTelemetry {
-		// To prevent leaking test data to prod,
-		// the prod TelemetryURL is set for the release builds during the build time.
-		// The dev TelemetryURL is set only when running `make run-debug`.
-		if c.TelemetryURL != "" {
-			go func() {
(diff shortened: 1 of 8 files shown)


recharte (commented):
Overall the code looks good but I'm not entirely sure that having a table called settings with a single ID column is the best long term solution.
Maybe we should go with a key value approach that will not only cover our current use-case of storing the unique everest instance ID but will allow us to expand it to any other settings that we need to store like default values for a DB engine configuration or whatever.
What do you think about this?

api/everest.go
```diff
@@ -65,6 +66,31 @@ func NewEverestServer(c *config.EverestConfig, l *zap.SugaredLogger) (*EverestSe
...
+		return e, err
+	}
+
+	if !c.DisableTelemetry {
+		// To prevent leaking test data to prod,
+		// the prod TelemetryURL is set for the release builds during the build time.
+		// The dev TelemetryURL is set only when running `make run-debug`.
+		if c.TelemetryURL != "" {
+			go func() {
+				e.runTelemetryJob(ctx, c)
```
michal-kralik, line 88: Can we hook this up to some global context which is cancelled with `ctrl + c`?

api/telemetry.go
```diff
@@ -0,0 +1,144 @@
...
+		return err
+	}
+	defer resp.Body.Close() //nolint:errcheck
+	if resp.StatusCode != http.StatusOK {
+		e.l.Info("Telemetry service responded with http status ", resp.StatusCode)
+	}
+	return nil
+}
+
+func (e *EverestServer) runTelemetryJob(ctx context.Context, c *config.EverestConfig) {
```
michal-kralik, line 68: Would it make sense to consider using an existing module, e.g. from PMM for this? If such thing exists of course...
It may have more advanced logic, e.g. retries and stuff.

Just an idea.

api/everest.go
```diff
@@ -65,6 +66,31 @@ func NewEverestServer(c *config.EverestConfig, l *zap.SugaredLogger) (*EverestSe
...
 	}
 	err := e.initEverest()
+	if err != nil {
+		e.l.Error(err)
+		return e, err
+	}
+
+	ctx := context.Background()
+
+	_, err = e.storage.CreateSettings(ctx, model.SettingsParams{ID: uuid.NewString()})
```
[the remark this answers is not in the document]
oksana-grishchenko: You're right, thanks. The new implementation sets the ID only once

model/settings.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// GetSettings returns Everest settings.
+func (db *Database) GetSettings(_ context.Context) (*Settings, error) {
+	var settings []Settings
+	err := db.gormDB.First(&settings).Error
+	if err != nil {
+		return nil, err
+	}
+	if len(settings) > 1 {
```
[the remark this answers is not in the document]
oksana-grishchenko: right. The updated version uses another approach, so the code is deleted

oksana-grishchenko:
> Maybe we should go with a key value approach that will not only cover our current use-case of storing the unique everest instance ID but will allow us to expand it to any other settings that we need to store like default values for a DB engine configuration or whatever.

@recharte thanks for the suggestion, the key-value approach sounds great. Updated, PTAL

recharte (approved):
Looks great :muscle:

api/everest.go
recharte: I have just approved the PR before looking at this comment, this actually a great idea to make sure we can properly shutdown the server.

api/telemetry.go
oksana-grishchenko: The initial idea was to start with something simple. The [Future improvements](https://www.notion.so/percona/Product-adoption-usage-metrics-c523d843a5f349dcb8905bbfdcd35f61?pvs=4#6b813516b07d4c63bacff3afecef43a2) section of the docs mentions that retry etc more sophisticated way to handle metrics. I think we could move it out of scope for now

recharte: I have no idea if such a module exists :shrug:

api/everest.go
oksana-grishchenko: That's a good point, thanks. Figuring it out, that's a tricky one

oksana-grishchenko: Moved the telemetry run to the main.go so we could cancel the context it after the os.Signal received, PTAL

api/telemetry.go
```diff
@@ -0,0 +1,147 @@
...
+func (e *EverestServer) RunTelemetryJob(ctx context.Context, c *config.EverestConfig) {
+	e.l.Debug("Starting background jobs runner.")
+
+	interval, err := time.ParseDuration(c.TelemetryInterval)
+	if err != nil {
+		e.l.Error(errors.Join(err, errors.New("could not parse telemetry interval")))
+		return
+	}
+
+	timer := time.NewTimer(0)
```
michal-kralik, line 78: Does setting the timer to `0` mean it will be triggered immediately?
If yes, we will send metrics on every restart. Is this intended?

Would it make sense to have a random delay, e.g. 5-10 mins after start so we don't flood our servers during too many restarts?

model/settings.go
```diff
@@ -0,0 +1,72 @@
...
+	ID    string
+	Key   string
+	Value string
+}
+
+// SettingParams represents params for Everest settings.
+type SettingParams struct {
+	Key   string
+	Value string
+}
```
michal-kralik, line 40: This seems to be unused?

api/telemetry.go
oksana-grishchenko: > Does setting the timer to 0 mean it will be triggered immediately?

It does. I think you're right, in case of the server fails to start at some point, the restarts may spam the telemetry. Added 5 min delay for the initial collecting

model/settings.go
oksana-grishchenko: thanks, removed it since there is no need for it currently

gen1us2k (approved):

michal-kralik (approved):

PR #277, opened 2023-10-25
EVEREST-574 Sunset of postgres for internal storage

gen1us2k:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-0

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**


api/telemetry.go
```diff
@@ -97,28 +97,29 @@ func (e *EverestServer) RunTelemetryJob(ctx context.Context, c *config.EverestCo
 }
 
 func (e *EverestServer) collectMetrics(ctx context.Context, url string) error {
-	everestID, err := e.storage.GetEverestID(ctx)
```
gen1us2k, line 100: @oksana-grishchenko What can we do here?

gen1us2k: As I understand it generates uuid and stores it in the database. Can we change to md5(gethostname())?

recharte: We can keep using the UUID and store it in a secret.

gen1us2k: Or we can store it in a secret

oksana-grishchenko: basically if we won't have the multi k8s support I don't see why would we need to generate a new identifier. We may use smth that already exists and wouldn't change as long as the Everest deployment is exists, for example the `uid` of the namespace the Everest is deployed into.

api/backup_storage.go
```diff
@@ -67,398 +62,138 @@ func (e *EverestServer) CreateBackupStorage(ctx echo.Context) error {
 	}
 
 	c := ctx.Request().Context()
-
-	existingStorage, err := e.storage.GetBackupStorage(c, nil, params.Name)
-	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+	s, err := e.kubeClient.GetBackupStorage(c, params.Name)
+	if err != nil {
```
oksana-grishchenko, line 66: if the "not found" case also appears as an error, we should handle it here as well

api/backup_storage.go
```diff
@@ -67,398 +62,138 @@ func (e *EverestServer) CreateBackupStorage(ctx echo.Context) error {
...
+			"AWS_SECRET_ACCESS_KEY": params.AccessKey,
+			"AWS_ACCESS_KEY_ID":     params.SecretKey,
+		},
+	})
 	if err != nil {
-		return ctx.JSON(http.StatusInternalServerError, Error{Message: pointer.ToString(err.Error())})
+
+		e.l.Error(err)
+		return ctx.JSON(http.StatusInternalServerError, Error{
+			Message: pointer.ToString("Failed to get BackupStorage"),
```
oksana-grishchenko, line 91: misleading message

api/backup_storage.go
```diff
@@ -67,398 +62,138 @@ func (e *EverestServer) CreateBackupStorage(ctx echo.Context) error {
...
-				return ctx.JSON(http.StatusBadRequest, Error{
-					Message: pointer.ToString("Backup storage with the same name already exists. " + pgErr.Detail),
-				})
-			}
-		}
 
 		e.l.Error(err)
 		return ctx.JSON(http.StatusInternalServerError, Error{
-			Message: pointer.ToString("Could not create a new backup storage"),
+			Message: pointer.ToString("Failed to get BackupStorage"),
```
oksana-grishchenko, line 110: misleading message

api/backup_storage.go
```diff
@@ -67,398 +62,138 @@ func (e *EverestServer) CreateBackupStorage(ctx echo.Context) error {
...
-		return ctx.JSON(http.StatusInternalServerError, Error{Message: pointer.ToString(err.Error())})
-	}
-	err = e.deleteBackupStorage(c, bs)
-
-	if err != nil {
+	if err := e.kubeClient.DeleteSecret(ctx.Request().Context(), fmt.Sprintf("%s-secret", backupStorageName)); err != nil {
+		e.l.Error(err)
 		return ctx.JSON(http.StatusInternalServerError, Error{
-			Message: pointer.ToString(err.Error()),
+			Message: pointer.ToString("Failed to get BackupStorage"),
```
oksana-grishchenko, line 136: misleading message

api/database_cluster.go
```diff
@@ -43,37 +39,6 @@ func (e *EverestServer) CreateDatabaseCluster(ctx echo.Context, kubernetesID str
...
-		}
-
-		err = kubeClient.EnsureConfigExists(ctx.Request().Context(), i, e.secretsStorage.GetSecret)
-		if err != nil {
-			e.l.Error(err)
-			return ctx.JSON(http.StatusBadRequest, Error{
-				Message: pointer.ToString("Could not create monitoring config in Kubernetes"),
-			})
-		}
-	}
```
oksana-grishchenko, line 75: perhaps we still should ensure that the mentioned BackupStorages and MonitoringConfig exist. From the other hand, this misconfiguration could happen only when using the pure API, on the FE there is no option to configure a custom name, you may only pick an existing one. So maybe we could get rid of it indeed

oksana-grishchenko (commented):
checked two files so far

api/backup_storage.go
api/database_cluster.go

love how many things we can just delete if we don't consider that multi-k8s thing❤️

One thing I'd like to highlight - **please ensure the changed API methods work at least once before merging the change to main**

api/backup_storage.go
gen1us2k: I just copy-pasted if err != nil handle and I'll revisit errors later =)

api/database_cluster.go
maxkondr: what about API first approach? In such a case API users may request invalid configuration and we do not try to validate it, right?

maxkondr: removing this part of code has nothing in common with removing PG

api/database_cluster.go
```diff
@@ -84,40 +49,7 @@ func (e *EverestServer) ListDatabaseClusters(ctx echo.Context, kubernetesID stri
...
-	names := kubernetes.BackupStorageNamesFromDBCluster(db)
-	e.waitGroup.Add(1)
-	go e.deleteK8SBackupStorages(context.Background(), kubeClient, names)
-
-	if db.Spec.Monitoring != nil && db.Spec.Monitoring.MonitoringConfigName != "" {
-		e.waitGroup.Add(1)
-		go e.deleteK8SMonitoringConfig(context.Background(), kubeClient, db.Spec.Monitoring.MonitoringConfigName)
-	}
-
-	return nil
```
maxkondr, line 120: this removed part is not related to PG removing

api/everest.go
```diff
@@ -183,17 +124,6 @@ func (e *EverestServer) Shutdown(ctx context.Context) error {
 	e.l.Info("Shutting down Everest")
 	e.waitGroup.Wait()
```
maxkondr, line 125: what does this waitGroup wait for?

api/everest.go
```diff
@@ -183,17 +124,6 @@ func (e *EverestServer) Shutdown(ctx context.Context) error {
...
-		if err := e.storage.Close(); err != nil {
-			e.l.Error(errors.Join(err, errors.New("could not shut down database storage")))
-		} else {
-			e.l.Info("Database storage shut down")
-		}
-	}()
-
 	done := make(chan struct{}, 1)
 	go func() {
 		e.waitGroup.Wait()
```
maxkondr, line 129: what does this waitGroup wait for?

pkg/kubernetes/backup_storage.go
```diff
@@ -1,64 +1,50 @@
...
-			return true, nil
-		}
-	}
+// ListBackupStorages returns list of managed database clusters.
+func (k *Kubernetes) ListBackupStorages(ctx context.Context) (*everestv1alpha1.BackupStorageList, error) {
+	return k.client.ListBackupStorages(ctx)
+}
 
-	return false, nil
+// GetBackupStorage returns database clusters by provided name.
```
maxkondr, line 32: comment is not related to function

maxkondr: and the rest in this file

api/backup_storage.go
```diff
@@ -17,448 +17,199 @@
...
-		if errors.As(err, &pgErr) {
-			if pgErr.Code.Name() == pgErrUniqueViolation {
-				return ctx.JSON(http.StatusBadRequest, Error{
-					Message: pointer.ToString("Backup storage with the same name already exists. " + pgErr.Detail),
-				})
-			}
-		}
-
 		e.l.Error(err)
 		return ctx.JSON(http.StatusInternalServerError, Error{
```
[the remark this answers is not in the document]
gen1us2k: No need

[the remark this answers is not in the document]
gen1us2k: Or it can be tricky pretty tricky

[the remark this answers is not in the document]
gen1us2k: To elaborate. The main idea is to validate everything before the creation of the backup storage to prevent 400xx errors in that case, if we received the error it means that it's either 5xx or cluster unavailable stuff and in that case, deletion of the secret will fail

deploy/quickstart-k8s.yaml
```diff
@@ -1,9 +1,54 @@
...
+    verbs: ["*"]
+  - apiGroups: [""]
+    resources: ["namespaces", "secrets"]
+    verbs: ["*"]
+  - apiGroups: [""]
+    resources: ["nodes", "pods", "persistentvolumes"]
+    verbs: ["get", "list"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["list"]
```
recharte, line 25: These need to be cluster role rules
```suggestion
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: everest-admin-cluster-role
rules:
  - apiGroups: [""]
    resources: ["nodes", "pods", "persistentvolumes"]
    verbs: ["get", "list"]
  - apiGroups: ["storage.k8s.io"]
    resources: ["storageclasses"]
    verbs: ["list"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: everest-admin-cluster-role-binding
roleRef:
  kind: "ClusterRole"
  apiGroup: "rbac.authorization.k8s.io"
  name: everest-admin-cluster-role
subjects:
  - kind: "ServiceAccount"
    name: everest-admin
```

api/kubernetes.go
```diff
@@ -17,143 +17,53 @@
...
-		Name:      k.Name,
-		Namespace: k.Namespace,
-		Uid:       k.UID,
+		Id:        "id",
+		Name:      "name",
+		Namespace: "namespace",
+		Uid:       "uid",
 	}
 	return ctx.JSON(http.StatusOK, result)
 }
```
recharte, line 56: Should we drop these API endpoints altogether? I know it breaks backward compatibility and it would require changes in the frontend as well as pretty much all endpoints that currently receive the kubernetes ID but I don't think it makes any sense to keep it with this hard-coded implementation.

gen1us2k: This is out of scope for this PR and it will be the next step to redesign API endpoints and have this done together with FE engineers

recharte: Makes sense

.github/workflows/ci.yml
```diff
@@ -245,28 +242,77 @@ jobs:
...
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+
 
       - name: Checkout CLI repo
         uses: actions/checkout@v4
         with:
           repository: percona/percona-everest-cli
-          ref: 'main'
+          ref: 'provision'
```
recharte, line 282: Leaving a comment so we don't forget to change this back before merge.

recharte (changes requested):
I have reviewed everything except for the monitoring config parts as it's not functional yet.
I tested the rest of the functionalities and it seems to be working fine with one exception, the method to update a backup storage location is broken.
Also, we removed some validations related with the backup storage locations that I think should be revisited, see comments below.

@gen1us2k please let me know when I should review the monitoring config parts.

As a side-note, with this change the kubernetes endpoint makes no sense at all. Also, all endpoints that use the kubernetes ID as a prefix should be changed. Let's not forget to work on this on a follow-up PR.

api/everest.go
gen1us2k: Thanks for catching this. It was leftover from the PG code removal

api/backup_storage.go
```diff
@@ -23,442 +23,241 @@ import (
...
-					Message: pointer.ToString("Backup storage with the same name already exists. " + pgErr.Detail),
+		e.l.Error(err)
+		if k8serrors.IsConflict(err) {
+			dErr := e.kubeClient.DeleteSecret(c, fmt.Sprintf("%s-secret", params.Name))
+			if dErr != nil {
+				return ctx.JSON(http.StatusInternalServerError, Error{
+					Message: pointer.ToString("Failing cleaning up the secret because failed creating backup storage"),
 				})
 			}
 		}
```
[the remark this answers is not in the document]
gen1us2k: There might be the following cases 
1. 4xx error -- we need to drop the secret
2. 5xx error - not sure that we need to drop the secret because the k8s cluster might be unavailable
3. TImeout error -- we do not need to drop the secret because the request might fail 

Originally the idea was to handle 5xx and timeout errors for deleting the secret yet I'll take a closer look before preparing the PR for the review

api/backup_storage.go
```diff
@@ -23,442 +23,241 @@ import (
...
-		Name:        s.Name,
-		Region:      s.Region,
-		Url:         &s.URL,
-	}
-
-	return ctx.JSON(http.StatusOK, result)
+	return ctx.JSON(http.StatusOK, BackupStorage{
+		Type: BackupStorageType(s.Spec.Type),
+		Name: s.Name,
+		// Description: &s.Spec.Description, //FIXME
```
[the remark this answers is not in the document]
gen1us2k: Updated it

api/backup_storage.go
```diff
@@ -17,448 +17,199 @@
...
-		URL:         params.Url,
-		Region:      params.Region,
-		AccessKeyID: newAccessKeyID,
-		SecretKeyID: newSecretKeyID,
+			return ctx.JSON(http.StatusInternalServerError, Error{
+				Message: pointer.ToString("Failed to get BackupStorage"),
+			})
+		}
+	}
+	err = e.kubeClient.UpdateBackupStorage(c, &everestv1alpha1.BackupStorage{
```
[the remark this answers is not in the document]
gen1us2k: It will be updated and a database cluster will be reconciled

api/backup_storage.go
```diff
@@ -17,448 +17,199 @@
...
 		return ctx.JSON(http.StatusConflict, Error{Message: pointer.ToString(err.Error())})
 	}
-
-	var accessKeyID, secretKeyID *string
-	defer e.cleanUpNewSecretsOnUpdateError(err, accessKeyID, secretKeyID)
-
-	accessKeyID, secretKeyID, err = e.createSecrets(c, &params.AccessKey, &params.SecretKey)
+	_, err = e.kubeClient.CreateSecret(c, &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-secret", params.Name),
```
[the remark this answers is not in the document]
gen1us2k: Yep. We have it in validation functions

api/backup_storage.go
```diff
@@ -17,448 +17,199 @@
...
-		return ctx.JSON(http.StatusInternalServerError, Error{Message: pointer.ToString(err.Error())})
+		e.l.Error(err)
+		return ctx.JSON(http.StatusInternalServerError, Error{
+			Message: pointer.ToString("Failed creating the secret for the backup storage"),
+		})
 	}
-	s, err := e.createBackupStorage(c, params, accessKeyID, secretKeyID)
+	err = e.kubeClient.CreateBackupStorage(c, &everestv1alpha1.BackupStorage{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      params.Name,
```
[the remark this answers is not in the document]
gen1us2k: Yep. We have it in validation functions

pkg/pmm/apikey.go
```diff
@@ -23,10 +23,15 @@ import (
...
 	"net/http"
+
+	"github.com/percona/percona-everest-backend/cmd/config"
 )
 
 // CreatePMMApiKey creates a new API key in PMM by using the provided username and password.
 func CreatePMMApiKey(ctx context.Context, hostname, apiKeyName, user, password string) (string, error) {
+	if config.Debug {
+		return "test-api-key", nil
+	}
```
recharte, line 34: Won't this completely break the monitoring for debug builds?

api/backup_storage.go
```diff
@@ -181,7 +181,7 @@ func (e *EverestServer) GetBackupStorage(ctx echo.Context, backupStorageName str
 }
 
 // UpdateBackupStorage updates of the specified backup storage.
-func (e *EverestServer) UpdateBackupStorage(ctx echo.Context, backupStorageName string) error {
+func (e *EverestServer) UpdateBackupStorage(ctx echo.Context, backupStorageName string) error { //nolint:funlen //FIXME
```
recharte, line 201: FIXME

api/monitoring_instance.go
```diff
@@ -160,7 +160,7 @@ func (e *EverestServer) GetMonitoringInstance(ctx echo.Context, name string) err
 }
 
 // UpdateMonitoringInstance updates a monitoring instance based on the provided fields.
-func (e *EverestServer) UpdateMonitoringInstance(ctx echo.Context, name string) error {
+func (e *EverestServer) UpdateMonitoringInstance(ctx echo.Context, name string) error { //nolint:funlen //FIXME
```
recharte, line 163: FIXME

api/backup_storage.go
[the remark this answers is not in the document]
recharte: I understand that some error codes might mean that the deletion of the secret might fail but that doesn't mean we shouldn't try... Leaving a dangling secret doesn't feel right. Why would we keep a secret without the parent backup storage?

api/backup_storage.go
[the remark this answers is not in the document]
recharte: Still commented here though

pkg/kubernetes/client/customresources/monitoringconfigs.go
```diff
@@ -109,3 +118,19 @@ func (c *monitoringConfigClient) Get(
 		Into(result)
 	return result, err
 }
+
+// List retrieves backup storage list based on opts.
```
recharte, line 122:
```suggestion
// List retrieves a monitoring configs list based on opts.
```

pkg/kubernetes/client/customresources/monitoringconfigs.go
```diff
@@ -49,28 +55,31 @@ type monitoringConfigClient struct {
 	namespace  string
 }
 
-// List lists database clusters based on opts.
-func (c *monitoringConfigClient) List(ctx context.Context, opts metav1.ListOptions) (*everestv1alpha1.MonitoringConfigList, error) {
-	result := &everestv1alpha1.MonitoringConfigList{}
+// Create creates a resource.
```
recharte, line 58:
```suggestion
// Create creates a monitoring config.
```

api/monitoring_instance.go
```diff
@@ -17,291 +17,238 @@
...
+			})
+		}
 		e.l.Error(err)
-		return ctx.JSON(http.StatusNotFound, Error{
-			Message: pointer.ToString("Could not find monitoring instance"),
+		return ctx.JSON(http.StatusInternalServerError, Error{
+			Message: pointer.ToString("Failed getting monitoring config"),
 		})
 	}
+	_ = m
```
recharte, line 181: Why?

api/monitoring_instance.go
```diff
@@ -17,291 +17,238 @@
...
 
-	var apiKeyID *string
-	if params.Pmm != nil {
-		keyID, err := e.createAndStorePMMApiKey(
-			ctx.Request().Context(), i.Name,
-			params.Url, params.Pmm.ApiKey, params.Pmm.User, params.Pmm.Password,
+	if params.Pmm != nil && params.Pmm.User != "" && params.Pmm.Password != "" {
```
recharte, line 183: This logic is different from the Create function, should it? Why can't we use an API key from params?

api/monitoring_instance.go
```diff
@@ -17,291 +17,238 @@
...
-		return kubernetes.IsMonitoringConfigInUse(ctx, name, kubeClient)
-	})
-	if err != nil && !errors.Is(err, kubernetes.ErrConfigInUse) {
-		e.l.Error(errors.Join(err, errors.New("could not delete monitoring config from kubernetes cluster")))
-		if errors.Is(err, kubernetes.ErrConfigInUse) {
-			return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Could not delete monitoring config from the Kubernetes cluster")})
+		if err != nil {
+			e.l.Error(err)
+			return ctx.JSON(http.StatusInternalServerError, Error{
+				Message: pointer.ToString("Could not create an API key in PMM"),
```
recharte, line 207: Incorrect error message.

api/monitoring_instance.go
```diff
@@ -17,291 +17,238 @@
...
-		apiKey, err = pmm.CreatePMMApiKey(
-			ctx, url, fmt.Sprintf("everest-%s-%s", name, uuid.NewString()),
-			user, password,
-		)
-		if err != nil {
-			e.l.Error(err)
-			return "", errors.New("could not create an API key in PMM")
+// DeleteMonitoringInstance deletes a monitoring instance.
+func (e *EverestServer) DeleteMonitoringInstance(ctx echo.Context, name string) error {
+	if err := e.kubeClient.DeleteMonitoringConfig(ctx.Request().Context(), name); err != nil {
```
recharte, line 248: Shouldn't we check if the monitoring config is in use by some DBC and error?

recharte (changes requested):

pkg/kubernetes/monitoring_config.go
```diff
@@ -13,146 +13,64 @@
...
-	return res, nil
+// DeleteMonitoringConfig returns monitoring configs by provided name.
+func (k *Kubernetes) DeleteMonitoringConfig(ctx context.Context, name string) error {
+	return k.client.DeleteMonitoringConfig(ctx, name)
 }
 
-// IsMonitoringConfigInUse returns true if a monitoring config is in use
-// by the provided Kubernetes cluster.
-func IsMonitoringConfigInUse(ctx context.Context, name string, kubeClient *Kubernetes) (bool, error) {
-	inUse, err := kubeClient.isMonitoringConfigUsedByVMAgent(ctx, name)
```
recharte, line 76: Why did we drop this?

api/database_cluster.go
gen1us2k: Done

pkg/pmm/apikey.go
gen1us2k: Nope

api/backup_storage.go
gen1us2k: Yeah. Later :)

api/monitoring_instance.go
```diff
@@ -32,7 +32,7 @@ import (
 )
 
 // CreateMonitoringInstance creates a new monitoring instance.
-func (e *EverestServer) CreateMonitoringInstance(ctx echo.Context) error { //nolint:funlen //FIXME
+func (e *EverestServer) CreateMonitoringInstance(ctx echo.Context) error { //nolint:funlen,cyclop //FIXME
```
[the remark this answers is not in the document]
gen1us2k: Yeah. Later :)

pkg/kubernetes/monitoring_config.go
gen1us2k: Restored

api/backup_storage.go
```diff
@@ -23,442 +23,257 @@ import (
...
-					Message: pointer.ToString("Backup storage with the same name already exists. " + pgErr.Detail),
-				})
-			}
-		}
-
 		e.l.Error(err)
+		// TODO: Move this logic to the operator
+		dErr := e.kubeClient.DeleteSecret(c, fmt.Sprintf("%s-secret", params.Name))
+		if dErr != nil {
+			return ctx.JSON(http.StatusInternalServerError, Error{
```
michal-kralik, line 110: The secret will be hanging around k8s indefinitely in this case. We shall clean it up.

api/backup_storage.go
```diff
@@ -23,442 +23,257 @@ import (
...
 // DeleteBackupStorage deletes the specified backup storage.
 func (e *EverestServer) DeleteBackupStorage(ctx echo.Context, backupStorageName string) error {
-	c := ctx.Request().Context()
-	bs, err := e.storage.GetBackupStorage(c, nil, backupStorageName)
+	used, err := e.kubeClient.BackupStorageIsUsed(ctx.Request().Context(), backupStorageName)
 	if err != nil {
-		if errors.Is(err, gorm.ErrRecordNotFound) {
-			return ctx.JSON(http.StatusNotFound, Error{Message: pointer.ToString("Could not find backup storage")})
+		if k8serrors.IsNotFound(err) {
+			return ctx.JSON(http.StatusNotFound, Error{
```
michal-kralik, line 135: We can return a successful response here.
The API call is meant to delete a backup storage and it's not available in the k8s cluster - hence deleted.

api/backup_storage.go
```diff
@@ -23,442 +23,257 @@ import (
...
+			Message: pointer.ToString("Failed to get BackupStorage"),
+		})
 	}
-	err = e.deleteBackupStorage(c, bs)
-
-	if err != nil {
+	if err := e.kubeClient.DeleteSecret(ctx.Request().Context(), fmt.Sprintf("%s-secret", backupStorageName)); err != nil {
+		e.l.Error(err)
+		if k8serrors.IsNotFound(err) {
+			return ctx.JSON(http.StatusNotFound, Error{
```
michal-kralik, line 163: We can return a successful response here.
The API call is meant to delete a backup storage and the associated secret. None of the are available in the k8s cluster - hence deleted.

api/backup_storage.go
```diff
@@ -23,442 +23,257 @@ import (
...
+				Message: pointer.ToString("Secret is not found"),
+			})
+		}
 		return ctx.JSON(http.StatusInternalServerError, Error{
```
michal-kralik, line 167: Why do we return an error here?
What do we expect the user to do with the error?

We shall return a successful response, log this as a warning and retry later.

api/backup_storage.go
```diff
@@ -23,442 +23,257 @@ import (
...
+		e.l.Error(err)
+		return ctx.JSON(http.StatusInternalServerError, Error{
+			Message: pointer.ToString("Failed to get BackupStorage"),
+		})
 	}
-	err = e.deleteBackupStorage(c, bs)
-
-	if err != nil {
+	if err := e.kubeClient.DeleteSecret(ctx.Request().Context(), fmt.Sprintf("%s-secret", backupStorageName)); err != nil {
+		e.l.Error(err)
```
michal-kralik, line 161: We're logging an error also in case the secret is not found.
There's no need to log the error in this particular case.

.github/workflows/ci.yml
gen1us2k: I assume we can do this in the following PRs because right now the CLI changes to use a specific commit hash to provision stuff instead of using something pointing to the main

api/backup_storage.go
```diff
@@ -23,442 +23,257 @@ import (
...
-
-	existingStorage, err := e.storage.GetBackupStorage(c, nil, params.Name)
-	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+	if code, err := e.checkBackupStorageExists(c, params); err != nil {
+		return ctx.JSON(code, Error{Message: pointer.ToString(err.Error())})
+	}
+	_, err = e.kubeClient.CreateSecret(c, &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-secret", params.Name),
+			Namespace: e.kubeClient.Namespace(),
```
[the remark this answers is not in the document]
gen1us2k: I kept the existing behavior. It's up to you to improve this later

api/backup_storage.go
```diff
@@ -23,442 +23,257 @@ import (
...
+		},
+		Type: corev1.SecretTypeOpaque,
+		StringData: map[string]string{
+			"AWS_SECRET_ACCESS_KEY": params.SecretKey,
+			"AWS_ACCESS_KEY_ID":     params.AccessKey,
+		},
+	})
+	if err != nil {
 		e.l.Error(err)
 		return ctx.JSON(http.StatusInternalServerError, Error{
```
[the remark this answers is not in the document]
gen1us2k: I'm not sure that we need to do that because this is looks like unexpected behavior to me

api/backup_storage.go
gen1us2k: Could you elaborate on what do you mean here? I'm not sure that I understand this. If BE failed to delete the secret, the end user will get the error message

api/backup_storage.go
gen1us2k: The tests provide a different picture and we have integration tests written for non-existing backup storage expecting a 404 error for the non-existing backup storage removal

api/backup_storage.go
gen1us2k: The tests provide a different picture and we have integration tests written for non-existing backup storage expecting a 404 error for the non-existing backup storage removal

api/backup_storage.go
gen1us2k: >We shall return a successful response, log this as a warning and retry later.

I had a conversation with the FE team about that topic and they need to implement this support on the FE side. Yet right now it will be clear to the end user that everest failed to perform a requested action with the correct statuses.

api/backup_storage.go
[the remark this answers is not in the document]
michal-kralik: Without this logic it will not be possible to create the backup storage.

api/backup_storage.go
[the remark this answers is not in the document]
michal-kralik: It's a small change. Let's do it in this PR.

api/backup_storage.go
michal-kralik: On line 110 the k8s resource state is:
- There is no backup storage config
- There is a hanging secret with credentials

This will prevent creating a new backup storage and we shall not leave the secret around if it's not needed.

api/backup_storage.go
michal-kralik: Yes, let's change that please.

api/backup_storage.go
michal-kralik: Yes, let's change that please.

api/backup_storage.go
michal-kralik: What do we need from the FE side?
This logic shall stay completely on the BE.

api/backup_storage.go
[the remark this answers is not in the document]
gen1us2k: Yet the end user will understand what went wrong. I do not understand why do we need to override anything

api/backup_storage.go
gen1us2k: This is unrelated to the changes in the PR. On the main branch the code has the same behavior

api/backup_storage.go
gen1us2k: Why? Don't understand how it will improve the UX or anything else taking into account that we have the same behavior right now

api/backup_storage.go
gen1us2k: Why? Don't understand how it will improve the UX or anything else taking into account that we have the same behavior right now

api/backup_storage.go
gen1us2k: How BE should return a warning with the retry to the end user?

michal-kralik: The end user shall not see the error. It shall be in the BE logs.
The end user shall not be concerned with this.

api/backup_storage.go
michal-kralik: There's nothing wrong when we reach line 163. There's no reason to return an error. The user cannot do anything about it.

api/backup_storage.go
michal-kralik: This PR is meant to improve our current situation, not replicate issues we have in the current setup.
One of the reasons you proposed removing PG was to resolve this type of issues which you mentioned will not be present in the new setup without PG.

api/backup_storage.go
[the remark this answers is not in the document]
michal-kralik: The end user cannot do anything to resolve the issue.

oksana-grishchenko (approved):
Re-reviewed, looks good, thanks for this massive effort! Left some minor comments. And let's of course have all-green pipelines back before merging

api/backup_storage.go
gen1us2k: They can use kubectl and everest uses kubernetes design using the implemented solution. It returns the same error for the end user as we had it before meaning that the resource does not exist and follows REST design principles and semantics of [HTTP 1.1](https://www.rfc-editor.org/rfc/rfc7231#section-4.3.5).

[RFC 7231, section 6.5.4 404 Not Found](https://www.rfc-editor.org/rfc/rfc7231#section-6.5.4) says 404 response is for a resource does not exist. Since no specific http method (in particular, not DELETE) is being called out to be treated otherwise, we can intuitively get an impression (and rightfully so), that the request DELETE /backup-storage should result in a 404.

The end user can use kubectl to understand why it does not exist, check secrets, and decide what to do next. Also, it can be used in the automation.

gen1us2k: and also the main goal for the changes in the PR is to keep behavior the same

api/backup_storage.go
gen1us2k: Let's continue the discussion here https://github.com/percona/percona-everest-backend/pull/277/files#r1384879742

api/backup_storage.go
```diff
@@ -23,442 +25,257 @@ import (
...
-		newSecretKeyID = &newID
-
-		// create new SecretKey
-		err := e.secretsStorage.PutSecret(ctx, newID, *secretKey)
-		if err != nil {
-			e.l.Error(err)
-			return newAccessKeyID, newSecretKeyID, errors.New("could not store secret key in secrets storage")
-		}
+	if params.BucketName != nil {
+		bs.Spec.Bucket = *params.BucketName
```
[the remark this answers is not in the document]
gen1us2k: Yep. Me too but we defining what should be updated right now. @recharte had a conversation about the topic and we can address it in the upcoming PRs.

api/backup_storage.go
```diff
@@ -23,442 +25,257 @@ import (
...
-
-func (e *EverestServer) deleteOldSecretsAfterUpdate(ctx context.Context, params *UpdateBackupStorageParams, s *model.BackupStorage) {
-	// delete old AccessKey
-	if params.AccessKey != nil {
-		cErr := e.secretsStorage.DeleteSecret(ctx, s.AccessKeyID)
-		if cErr != nil {
-			e.l.Errorf("Failed to delete unused secret, please delete it manually. id = %s", s.AccessKeyID)
-		}
+	if params.Region != nil {
+		bs.Spec.Bucket = *params.Region
```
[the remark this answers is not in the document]
gen1us2k: Fixed! Thanks for catching this

api/backup_storage.go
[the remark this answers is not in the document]
gen1us2k: No, because we need to provide backward compatibility because updating backup storages will fail after the upgrade. Let's keep it as is

pkg/kubernetes/backup_storage.go
```diff
@@ -1,64 +1,92 @@
...
-		names := BackupStorageNamesFromDBCluster(&db)
-		if _, ok := names[name]; ok {
-			return true, nil
-		}
+	options := metav1.ListOptions{
+		LabelSelector: metav1.FormatLabelSelector(&metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				fmt.Sprintf(backupStorageNameLabelTmpl, backupStorageName): backupStorageLabelValue,
+			},
+		}),
```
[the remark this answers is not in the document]
gen1us2k: We use the same approach for listing backups/restores per database and we have those labels added during reconciliation by operator

pkg/kubernetes/kubernetes.go
```diff
@@ -45,54 +44,48 @@ const (
...
-type secretGetter interface {
-	GetSecret(ctx context.Context, id string) (string, error)
-}
-
-// New returns new Kubernetes object.
-func New(kubeconfig []byte, namespace string, l *zap.SugaredLogger) (*Kubernetes, error) {
-	client, err := client.NewFromKubeConfig(kubeconfig, namespace)
+// NewInCluster creates a new kubernetes client using incluster authentication.
+func NewInCluster(l *zap.SugaredLogger) (*Kubernetes, error) {
+	client, err := client.NewIncluster()
```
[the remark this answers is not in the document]
gen1us2k: Done

pkg/kubernetes/client/monitoring_config.go
```diff
@@ -7,23 +23,29 @@ import (
...
 	return err
 }
 
-// GetMonitoringConfig returns the MonitoringConfig.
-func (c *Client) GetMonitoringConfig(ctx context.Context, name string) (*everestv1alpha1.MonitoringConfig, error) {
-	return c.customClientSet.MonitoringConfig(c.namespace).Get(ctx, name, metav1.GetOptions{})
+// UpdateMonitoringConfig updates an monitoringConfig.
+func (c *Client) UpdateMonitoringConfig(ctx context.Context, storage *everestv1alpha1.MonitoringConfig) error {
+	_, err := c.customClientSet.MonitoringConfig(storage.Namespace).Update(ctx, storage, metav1.UpdateOptions{})
+	return err
```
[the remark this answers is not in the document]
gen1us2k: Done

pkg/kubernetes/client/monitoring_config.go
```diff
@@ -7,23 +23,29 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// CreateMonitoringConfig creates an MonitoringConfig.
-func (c *Client) CreateMonitoringConfig(ctx context.Context, mc *everestv1alpha1.MonitoringConfig) error {
-	_, err := c.customClientSet.MonitoringConfig(c.namespace).Post(ctx, mc, metav1.CreateOptions{})
+// CreateMonitoringConfig creates an monitoringConfig.
+func (c *Client) CreateMonitoringConfig(ctx context.Context, storage *everestv1alpha1.MonitoringConfig) error {
+	_, err := c.customClientSet.MonitoringConfig(storage.Namespace).Create(ctx, storage, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
gen1us2k: Fixed

api/backup_storage.go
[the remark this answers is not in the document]
recharte: > we need to provide backward compatibility
we don't, we agreed that while in alpha it's fine to break backward compatibility

api/backup_storage.go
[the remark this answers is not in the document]
recharte: @michal-kralik this is a very valid concern. However I'm not sure what's the best way of handling it.
1. Update/recreate the secret:
* Pro: No action required by the user, we automatically fix the issue and proceed with creating the backup storage
* Con: We can't ensure we aren't breaking something that uses the secret we're overwriting
2. Fail and let the user know:
* Pro: We ensure we don't fuck things up
* Con: The user needs to manually delete the secret using kubectl.

Our current implementation of 2. is flawed because we don't properly tell the user what happened so he won't be able to act to fix it.

api/backup_storage.go
recharte: I'm pretty sure that the widely accepted expectation for REST APIs is that the DELETE call shall be idempotent in terms of the state of the server but not in the status code returned. The 2nd (and 3rd, 4th, ...) DELETE calls should return a 404.
I don't think we should apply this change.

api/backup_storage.go
recharte: If in https://github.com/percona/percona-everest-backend/pull/277/files#r1386701654 I suggested to return 404, here I'm not so sure. We did delete the backup storage successfully. I feel that if after deleting the backup storage we can't find the secret maybe we should state the operation as successful and return a 204.

api/backup_storage.go
recharte: When?

api/backup_storage.go
recharte: @michal-kralik Do you have a suggestion on how to approach this cleanup? I feel like we have no option other than log and fail. Maybe we do need to improve the error message to say exactly the name of the unbound secret so that the everest admin can actually act on this message. What do you think?

api/backup_storage.go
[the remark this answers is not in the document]
recharte: I think that it's fair to assume that only everest should manage the resources on it's namespace so the con of 1. can be removed. IMO, in order to reduce the amount of user interactions in error handling cases I think we should go with option 1.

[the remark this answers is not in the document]
gen1us2k: >Our current implementation of 2. is flawed because we don't properly tell the user what happened so he won't be able to act to fix it.
Agree and handled it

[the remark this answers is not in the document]
michal-kralik: @recharte I remember at some point we added a label `k8s.io/managed-by: percona-everest` to k8s resources. Or something along these lines.
We can check for this label and update the secret without the risk of breaking other stuff.

api/backup_storage.go
michal-kralik: @recharte the end goal would be we clean up resources which should have been deleted but were not due to some issue. There's a couple approaches to this. Too long for a GitHub comment, but happy to elaborate:

1. Automatically retry removal in the background until the backend process is restarted
2. Reconcile periodically and remove unused resources - e.g. if this secret is not used as it should, it's not needed
3. Keep a state and try cleaning up resources which we were not able to clean before, so we don't have to do this periodically
4. etc.

The easiest to implement is 1. It may be good enough.
If we're looking for a more robust solution we could go for 2. or 3.

api/backup_storage.go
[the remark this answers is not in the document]
recharte: option 1. implemented in 99c49783afa0546f03f7c03f3d0e4d1176cc9204

api/backup_storage.go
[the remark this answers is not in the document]
recharte: suggestion implemented in 3677adff8e3feec7f718280d4c68710567385888

pkg/kubernetes/client/customresources/backupstorages.go
```diff
@@ -13,6 +13,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package customresources ...
```
maxkondr, line 16: not an informative comment

pkg/pmm/apikey.go
maxkondr: so, in case developer wants to run Everest with debug enabled the interaction with PMM will be "emulated", right? this is not what is expected from enabling debug

api/validation.go
```diff
@@ -424,44 +401,30 @@ func validateCreateDatabaseClusterRequest(dbc DatabaseCluster) error {
...
 	}
-	engine, err := kubeClient.GetDatabaseEngine(ctx.Request().Context(), engineName)
+	engine, err := e.kubeClient.GetDatabaseEngine(ctx.Request().Context(), engineName)
 	if err != nil {
 		return err
 	}
 	if err := validateVersion(databaseCluster.Spec.Engine.Version, engine); err != nil {
 		return err
 	}
+	if databaseCluster.Spec.Monitoring != nil && databaseCluster.Spec.Monitoring.MonitoringConfigName != nil {
```
michal-kralik, line 420: We shall validate somewhere if `databaseCluster.Spec != nil`

pkg/kubernetes/backup_storage.go
```diff
@@ -1,64 +1,92 @@
...
-	for _, db := range dbs.Items {
-		db := db
-		names := BackupStorageNamesFromDBCluster(&db)
-		if _, ok := names[name]; ok {
-			return true, nil
-		}
+	options := metav1.ListOptions{
+		LabelSelector: metav1.FormatLabelSelector(&metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				fmt.Sprintf(backupStorageNameLabelTmpl, backupStorageName): backupStorageLabelValue,
```
michal-kralik, line 66: With this approach, I believe we're opening up room for errors.
- Whenever, in the future, anyone forgets to add a label, this code does not detect it.
- The approach allows for race-conditions. What if the operator has not yet added the label?

What do we aim to gain by using this approach compared to just listing all db clusters and other resources?

pkg/kubernetes/backup_storage.go
```diff
@@ -1,64 +1,92 @@
...
-	for _, b := range backups.Items {
-		if b.Spec.BackupStorageName == name {
-			return true, nil
-		}
+	if len(list.Items) > 0 {
+		return true, nil
 	}
-
-	restores, err := kubeClient.ListDatabaseClusterRestores(ctx)
+	bList, err := k.client.ListDatabaseClusterBackups(ctx, options)
```
michal-kralik, line 77: Do PG backups created from a schedule properly reference the backup storage via labels?

api/database_cluster.go
```diff
@@ -168,43 +89,17 @@ func (e *EverestServer) UpdateDatabaseCluster(ctx echo.Context, kubernetesID str
 	if ctx.Response().Status >= http.StatusMultipleChoices {
 		return nil
 	}
```
michal-kralik, line 91: This code is no longer needed

api/everest.go
```diff
@@ -20,110 +20,48 @@ package api
...
+		l:          l,
+		echo:       echo.New(),
+		kubeClient: kubeClient,
 	}
 	if err := e.initHTTPServer(); err != nil {
 		return e, err
 	}
-	err := e.initEverest(context.Background())
-	if err != nil {
-		e.l.Error(err)
```
michal-kralik, line 70: What is the reason we dropped all error logging in this function?
It will be hard to debug problems during start of Everest.

api/database_cluster_backup.go
```diff
@@ -48,76 +47,23 @@ func (e *EverestServer) ListDatabaseClusterBackups(ctx echo.Context, kubernetesI
...
-				Message: pointer.ToString("Could not create BackupStorage"),
-			})
-		}
-	}
-
-	if err := e.validateDBClusterAccess(ctx, kubernetesID, backup.Spec.DbClusterName); err != nil {
-		return err
+	if err := validateDatabaseClusterBackup(ctx.Request().Context(), dbb, e.kubeClient); err != nil {
+		e.l.Error(err)
+		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString(err.Error())})
```
michal-kralik, line 59: We shall not return the error in the message.
`validateDatabaseClusterBackup` returns internal error messages.
It shall either return external error messages only, or we need to have additional logic here to figure out what the error message displayed to the user will be.

api/validation.go
```diff
@@ -639,3 +614,86 @@ func validateDatabaseClusterOnUpdate(dbc *DatabaseCluster, oldDB *everestv1alpha
...
 }
+
+func validateDatabaseClusterBackup(ctx context.Context, backup *DatabaseClusterBackup, kubeClient *kubernetes.Kubernetes) error {
+	if backup == nil {
+		return errors.New("backup cannot be empty")
+	}
+	if backup.Spec == nil {
+		return errors.New(".spec cannot be empty")
+	}
+	b := &everestv1alpha1.DatabaseClusterBackup{}
```
michal-kralik, line 625: Why do we need `everestv1alpha1.DatabaseClusterBackup`?
We have all data required for validation in `backup` variable already, right?

api/kubernetes.go
```diff
@@ -17,188 +17,74 @@
...
 	corev1 "k8s.io/api/core/v1"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/client-go/tools/clientcmd"
 
-	"github.com/percona/percona-everest-backend/model"
 	"github.com/percona/percona-everest-backend/pkg/kubernetes"
 )
 
 // ListKubernetesClusters returns list of k8s clusters.
 func (e *EverestServer) ListKubernetesClusters(ctx echo.Context) error {
```
michal-kralik, line 31: What is the purpose of this function?

api/kubernetes.go
```diff
@@ -17,188 +17,74 @@
...
+			Id:        "id",
+			Name:      "name",
+			Namespace: "namespace",
+			Uid:       "uid",
+		},
+	})
 }
 
 // RegisterKubernetesCluster registers a k8s cluster in Everest server.
 func (e *EverestServer) RegisterKubernetesCluster(ctx echo.Context) error {
```
michal-kralik, line 43: Why do we have this function?

api/kubernetes.go
```diff
@@ -17,188 +17,74 @@
...
 }
 
 // GetKubernetesCluster Get the specified Kubernetes cluster.
-func (e *EverestServer) GetKubernetesCluster(ctx echo.Context, kubernetesID string) error {
-	k, err := e.storage.GetKubernetesCluster(ctx.Request().Context(), kubernetesID)
-	if err != nil {
-		e.l.Error(err)
-		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString("Could not find Kubernetes cluster")})
-	}
+func (e *EverestServer) GetKubernetesCluster(ctx echo.Context, _ string) error {
```
michal-kralik, line 48: Same here. What is the purpose of this function?

api/kubernetes.go
```diff
@@ -17,188 +17,74 @@
...
+		Name:      "name",
+		Namespace: "namespace",
+		Uid:       "uid",
 	}
 	return ctx.JSON(http.StatusOK, result)
 }
 
 // UnregisterKubernetesCluster removes a Kubernetes cluster from Everest.
-func (e *EverestServer) UnregisterKubernetesCluster(ctx echo.Context, kubernetesID string) error {
+func (e *EverestServer) UnregisterKubernetesCluster(ctx echo.Context, _ string) error {
```
michal-kralik, line 59: Why do we need API for unregistration?
It's not like we can unregister a cluster with this approach, right? :)

michal-kralik (commented):
Can we please extend this PR with information how to setup a local development environment so it's clear what the workflow for developing new features is?

api/kubernetes.go
```diff
@@ -234,82 +120,32 @@ func (e *EverestServer) SetKubernetesClusterMonitoring(ctx echo.Context, kuberne
...
-		if k8serrors.IsNotFound(err) {
-			// Nothing to disable
-			return ctx.NoContent(http.StatusOK)
-		}
-
-		e.l.Error(err)
-		return ctx.JSON(http.StatusInternalServerError, Error{
-			Message: pointer.ToString("Could not get VMAgent from Kubernetes"),
-		})
-	}
```
michal-kralik, line 261: What is the reason we dropped this part of code responsible for checking if VMAgent exists?
It ensured you don't get an error message if the VMAgent is not found.

In such case, the new code returns `Could not delete VMAgent` even though that's not true since there's no VMAgent running.

api/backup_storage.go
oksana-grishchenko: I think if the BackupStorage is deleted successfully, the method should return successful response. A failure in the secret deletion shouldn't be a blocker, it should be enough just to log it. 

> we have integration tests written for non-existing backup storage expecting a 404 error 

Don't see any problem here, if the _storage_ doesn't exist, the method should return 404. 

If the BE was acting like this before, we may keep it as it is bc it's out of scope for this PR. But imo it works wrong and if we keep it then let's create a tech debt ticket. Great that it was noticed and discussed.

.github/workflows/ci.yml
```diff
@@ -245,61 +244,127 @@ jobs:
...
+      - name: Expose local registry
         run: |
-          make local-env-up
-          make build-debug
-          SECRETS_ROOT_KEY=$(openssl rand -base64 32) make run-debug &> everest-backend.log &
+          kubectl port-forward --namespace kube-system service/registry 5000:80 &
+
+      - name: Build Everest backend
+        run: |
+          CGO_ENABLED=0 GOOS=linux GOARCH=amd64 make build-debug
```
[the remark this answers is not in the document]
gen1us2k: Because we reuse it for development and we have different CPU Arch and OS version

.github/workflows/ci.yml
```diff
@@ -245,61 +244,127 @@ jobs:
...
+          images: localhost:5000/perconalab/everest
+          tags:
+            dev-latest
+
+      - name: Build and Push everest dev-latest image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
```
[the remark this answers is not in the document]
gen1us2k: The first action sets tags for the image. The second action builds the image

.github/workflows/ci.yml
```diff
@@ -245,61 +244,127 @@ jobs:
...
         run: git config --global url."https://percona-platform-robot:${ROBOT_TOKEN}@github.com".insteadOf "https://github.com"
 
+      # We need to have Everest CRDs available before running provisioning and everest backend
+      # to have an ability to create monitoring configs and use them during the provisioning as
+      # a mock pmm server without running a real PMM.
+      - name: Install everest operator without Everest
+        run: |
+          cat <<EOF | kubectl apply -f -
+          apiVersion: v1
+          kind: Namespace
```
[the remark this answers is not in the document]
gen1us2k: Makes sense. Thanks for catching this. Updated

.github/workflows/ci.yml
```diff
@@ -245,61 +244,127 @@ jobs:
...
+              app.kubernetes.io/component: manager
+              app.kubernetes.io/created-by: everest-operator
+              app.kubernetes.io/instance: system
+              app.kubernetes.io/managed-by: kustomize
+              app.kubernetes.io/name: namespace
+              app.kubernetes.io/part-of: everest-operator
+              control-plane: controller-manager
+            name: percona-everest
+          EOF
+          curl https://raw.githubusercontent.com/percona/everest-operator/main/deploy/bundle.yaml -o bundle.yaml
```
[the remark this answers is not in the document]
gen1us2k: For this case, we just need the latest version of the operator

.github/workflows/ci.yml
```diff
@@ -245,61 +244,127 @@ jobs:
...
-              "type": "pmm",
-              "url": "http://localhost",
-              "name": "pmm-local",
-              "pmm": {
-                "apiKey": "dummy-key"
-              }
-            }'
-
-      - name: Run Provisioning
+          cat <<EOF | kubectl apply -f -
```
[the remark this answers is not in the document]
gen1us2k: Because we mock PMM here without credentials. In the other hand we need to deploy PMM

api-tests/tests/database-cluster-backups.spec.ts
```diff
@@ -78,14 +78,13 @@ test('dbcluster not found', async ({ request }) => {
     },
   }
 
-  let response = await request.post(`/v1/kubernetes/${kubernetesId}/database-cluster-backups`, {
+  const response = await request.post(`/v1/kubernetes/${kubernetesId}/database-cluster-backups`, {
```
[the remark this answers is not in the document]
gen1us2k: No. Just a small linter fixes

api/cluster-info.go
```diff
@@ -14,18 +14,13 @@ const (
...
 
 // GetKubernetesClusterInfo returns the cluster type and storage classes of a kubernetes cluster.
-func (e *EverestServer) GetKubernetesClusterInfo(ctx echo.Context, kubernetesID string) error {
-	_, kubeClient, code, err := e.initKubeClient(ctx.Request().Context(), kubernetesID)
-	if err != nil {
-		e.l.Error(err)
-		return ctx.JSON(code, Error{Message: pointer.ToString("Failed building connection to the Kubernetes cluster")})
-	}
-	clusterType, err := kubeClient.GetClusterType(ctx.Request().Context())
+func (e *EverestServer) GetKubernetesClusterInfo(ctx echo.Context, _ string) error {
```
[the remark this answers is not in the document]
gen1us2k: We will drop it in the next PR because we need to redesign this and have FE integrated

api/everest.go
gen1us2k: There is only initHTTPServer left and we log this error in main.go file

api/kubernetes.go
gen1us2k: Keep backward compatibility with the existing FE integration. We will drop this in the next PR addressing redesign of the API

api/kubernetes.go
gen1us2k: We will drop this in the next PR addressing redesign of the API

api/kubernetes.go
gen1us2k: We will drop this in the next PR addressing redesign of the API

api/validation.go
```diff
@@ -639,3 +614,86 @@ func validateDatabaseClusterOnUpdate(dbc *DatabaseCluster, oldDB *everestv1alpha
...
+	data, err := json.Marshal(backup)
+	if err != nil {
+		return err
+	}
+	if err := json.Unmarshal(data, b); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
gen1us2k: We receive because this spec is taken from everest-operator specification

api/monitoring_instance.go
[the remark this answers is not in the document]
gen1us2k: can you please elaborate of what's missing and where?

pkg/kubernetes/backup_storage.go
gen1us2k: What are your suggestions here?

pkg/kubernetes/backup_storage.go
gen1us2k: Yes

pkg/kubernetes/client/client.go
```diff
@@ -53,30 +53,27 @@ type Client struct {
...
-	}
-
 	config.QPS = defaultQPSLimit
 	config.Burst = defaultBurstLimit
 	config.Timeout = 10 * time.Second
 	clientset, err := kubernetes.NewForConfig(config)
 	if err != nil {
 		return nil, err
 	}
+	namespace, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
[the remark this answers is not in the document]
gen1us2k: We use it only once and this is only once appearance

pkg/pmm/apikey.go
gen1us2k: the Debug we have only for backups/pmm integrations and use it in CI only

api/cluster-info.go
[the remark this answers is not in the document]
michal-kralik: Can you please link the ticket where we have this documented?

pkg/kubernetes/backup_storage.go
michal-kralik: As mentioned in the original comment:
> listing all db clusters and other resources

api/everest.go
michal-kralik: There are other method calls, such as `kubernetes.NewInCluster()`

pkg/kubernetes/backup_storage.go
gen1us2k: I just keep everything consistent and in that case, we need to apply the same logic for listing backups/restores for a database cluster. It uses the same approach as we used before

api/kubernetes.go
recharte: https://jira.percona.com/browse/EVEREST-608

api/cluster-info.go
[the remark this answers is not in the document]
recharte: https://jira.percona.com/browse/EVEREST-608

pkg/pmm/apikey.go
maxkondr: in such a case this is not an appropriate way to use Debug

api/backup_storage.go
gen1us2k: The easiest solution does not work because this should not be the part of backend's logic. Instead, we should implement ownership across CR and use [finalizers](https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#external-resources) for MonitoringConfigs and BackupStorages.

In that case, we can cleanup everything for either secret removal or upstream CR removal. This is the common practice to use operator's pattern and recommended by operator-sdk

api-tests/tests/database-cluster-backups.spec.ts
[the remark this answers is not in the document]
maxkondr: why other linter issues are just silently resolved in such a case?

.github/workflows/ci.yml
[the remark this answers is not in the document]
recharte: Will be handled as part of https://jira.percona.com/browse/EVEREST-405

api/backup_storage.go
```diff
@@ -23,442 +25,277 @@ import (
...
+			"AWS_ACCESS_KEY_ID":     params.AccessKey,
+		},
+	})
 	if err != nil {
-		var pgErr *pq.Error
-		if errors.As(err, &pgErr) {
-			if pgErr.Code.Name() == pgErrUniqueViolation {
-				return ctx.JSON(http.StatusBadRequest, Error{
-					Message: pointer.ToString("Backup storage with the same name already exists. " + pgErr.Detail),
+		if k8serrors.IsAlreadyExists(err) {
```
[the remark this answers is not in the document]
gen1us2k: I have a PoG for this quarter to describe the architecture of everest and I'll write about relationships and who's responsible for handling the deletion/removal of entities we use. Is it be or operator.

api/backup_storage.go
```diff
@@ -23,442 +25,277 @@ import (
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      params.Name,
+					Namespace: e.kubeClient.Namespace(),
+				},
+				Type: corev1.SecretTypeOpaque,
+				StringData: map[string]string{
+					"AWS_SECRET_ACCESS_KEY": params.SecretKey,
+					"AWS_ACCESS_KEY_ID":     params.AccessKey,
+				},
+			})
```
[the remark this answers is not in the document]
gen1us2k: I'll create a Jira ticket to handle it

api/backup_storage.go
gen1us2k: We will address this in [this](https://jira.percona.com/browse/EVEREST-611) ticket

api/backup_storage.go
gen1us2k: We agreed that for backupstorages/monitoring configs objects we should return 404 but if we didn't find a secret we return 204

api/backup_storage.go
```diff
@@ -23,442 +25,277 @@ import (
...
+		return ctx.JSON(http.StatusInternalServerError, Error{
+			Message: pointer.ToString("Failed getting backup storage"),
 		})
 	}
-
-	var newAccessKeyID, newSecretKeyID *string
-	defer e.cleanUpNewSecretsOnUpdateError(err, newAccessKeyID, newSecretKeyID)
-
-	newAccessKeyID, newSecretKeyID, err = e.createSecrets(c, params.AccessKey, params.SecretKey)
+	params, err := validateUpdateBackupStorageRequest(ctx, bs, e.l)
```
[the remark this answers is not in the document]
gen1us2k: I agree that we need to have the same logic for update/create and I created a ticket to handle that https://jira.percona.com/browse/EVEREST-616

api/backup_storage.go
```diff
@@ -23,442 +25,277 @@ import (
...
+func (e *EverestServer) checkBackupStorageExists(c context.Context, params *CreateBackupStorageParams) (int, error) {
+	s, err := e.kubeClient.GetBackupStorage(c, params.Name)
+	if err != nil && !k8serrors.IsNotFound(err) {
 		e.l.Error(err)
-		return http.StatusInternalServerError, errors.New("could not update backup storage")
+		return http.StatusInternalServerError, errors.New("failed getting a backup storage from the Kubernetes cluster")
 	}
-
-	return 0, nil
+	if s != nil && s.Name != "" {
```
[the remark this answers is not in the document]
gen1us2k: The main problem right now is that the operator does not return `nil` structs because of the design.

api/backup_storage.go
```diff
@@ -23,442 +25,277 @@ import (
...
+		return http.StatusConflict, fmt.Errorf("storage %s already exists", params.Name)
```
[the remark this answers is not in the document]
gen1us2k: Thanks. Fixed

api/validation.go
```diff
@@ -424,44 +401,30 @@ func validateCreateDatabaseClusterRequest(dbc DatabaseCluster) error {
...
+	engine, err := e.kubeClient.GetDatabaseEngine(ctx.Request().Context(), engineName)
 	if err != nil {
 		return err
 	}
 	if err := validateVersion(databaseCluster.Spec.Engine.Version, engine); err != nil {
 		return err
 	}
+	if databaseCluster.Spec.Monitoring != nil && databaseCluster.Spec.Monitoring.MonitoringConfigName != nil {
+		if _, err := e.kubeClient.GetMonitoringConfig(context.Background(), *databaseCluster.Spec.Monitoring.MonitoringConfigName); err != nil {
+			if k8serrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
gen1us2k: Fixed. THanks

api/validation.go
```diff
@@ -470,12 +433,24 @@ func (e *EverestServer) validateDatabaseClusterCR(ctx echo.Context, kubernetesID
...
+		for _, schedule := range *databaseCluster.Spec.Backup.Schedules {
+			_, err := e.kubeClient.GetBackupStorage(context.Background(), schedule.BackupStorageName)
+			if err != nil {
+				if k8serrors.IsNotFound(err) {
+					return fmt.Errorf("backup storage %s does not exist", schedule.BackupStorageName)
+				}
+				return fmt.Errorf("could not validate backup storage %s", schedule.BackupStorageName)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
gen1us2k: because it requires k8s client and validateBackupSpec does not support it right now and it will be addressed here https://jira.percona.com/browse/EVEREST-616

api/validation.go
gen1us2k: We receive it because this spec is taken from everest-operator specification. `everestv1alpha1.DatabaseClusterBackup` Has defined types for each Field that we have and DatabaseBackup doesn't

pkg/pmm/apikey.go
gen1us2k: Yet we use it for the pipeline only

michal-kralik (approved):
Per discussion - there's a couple unresolved comments which we'll address in the near future.
We would like the QA to start testing as soon as possible and opted for merging early.

maxkondr (approved):

PR #303, opened 2023-11-10
EVEREST-107: config and secrets consistency

michal-kralik:
**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-107

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/api/backup_storage.go
+++ b/api/backup_storage.go
@@ -16,6 +16,7 @@
 package api
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 
@@ -47,6 +48,7 @@ func (e *EverestServer) ListBackupStorages(ctx echo.Context) error {
 			BucketName:  s.Spec.Bucket,
 			Region:      s.Spec.Region,
 			Url:         &s.Spec.EndpointURL,
+			Status:      s.Status.Status,
 		})
 	}
 
@@ -54,7 +56,7 @@ func (e *EverestServer) ListBackupStorages(ctx echo.Context) error {
 }
 
 // CreateBackupStorage creates a new backup storage object.
-func (e *EverestServer) CreateBackupStorage(ctx echo.Context) error { //nolint:funlen
+func (e *EverestServer) CreateBackupStorage(ctx echo.Context) error {
 	params, err := validateCreateBackupStorageRequest(ctx, e.l)
 	if err != nil {
 		return ctx.JSON(http.StatusBadRequest, Error{Message: pointer.ToString(err.Error())})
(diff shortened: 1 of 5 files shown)


api/backup_storage.go
```diff
@@ -110,36 +86,70 @@ func (e *EverestServer) CreateBackupStorage(ctx echo.Context) error { //nolint:f
 			Type:                  everestv1alpha1.BackupStorageType(params.Type),
 			Bucket:                params.BucketName,
 			Region:                params.Region,
-			CredentialsSecretName: params.Name,
+			CredentialsSecretName: "",
+		},
+		Status: {
+			Status: "Initializing",
```
recharte, line 92: I like the idea of using the status field to help us manage what is currently an inconsistent state and I think we can leverage this to achieve the data consistency we're looking for.
However, I don't like the fact that the backend is the entity responsible for setting this status.
For all intents and purposes, the backend is just a client of the everest CRDs and as such should only be responsible for setting their spec and never a status.

recharte:
> Create backup storage
>
> 1. Create backup storage in Initializing state
> 2. Create a secret with owner reference to backup storage
> 3. Update backup storage with secret name and set state to Ready

I like the idea of using the status field to help us manage what is currently an inconsistent state and I think we can leverage this to achieve the data consistency we're looking for.
However, I don't like the fact that the backend is the entity responsible for setting this status.
For all intents and purposes, the backend is just a client of the everest CRDs and as such should only be responsible for setting their spec and never a status.

api/backup_storage.go
```diff
@@ -117,36 +86,70 @@ func (e *EverestServer) CreateBackupStorage(ctx echo.Context) error { //nolint:f
...
+			e.l.Errorf("Could not delete backup storage %s in initializing state", bs.Name)
 		}
+
 		return ctx.JSON(http.StatusInternalServerError, Error{
-			Message: pointer.ToString("Failed creating backup storage"),
+			Message: pointer.ToString("Could not create a secret for the backup storage"),
 		})
 	}
+
+	bs.Status.Status = "Ready"
```
oksana-grishchenko, line 126: Let's have the possible statuses defined as constants in one place

api/backup_storage.go
```diff
@@ -117,36 +86,70 @@ func (e *EverestServer) CreateBackupStorage(ctx echo.Context) error { //nolint:f
...
-		BucketName:  params.BucketName,
-		Region:      params.Region,
-		Url:         params.Url,
+		Type:        BackupStorageType(bs.Spec.Type),
+		Name:        bs.Name,
+		Description: &bs.Spec.Description,
+		BucketName:  bs.Spec.Bucket,
+		Region:      bs.Spec.Region,
+		Url:         &bs.Spec.EndpointURL,
+		Status:      bs.Status.Status,
```
oksana-grishchenko, line 152: Do we need to expose the status to the API? Looks like the "Initializing" will never returned to the API user anyway.

PR #312, opened 2023-11-16
EVEREST-496 Multi namespace support

gen1us2k:
[EVEREST-496](https://jira.percona.com/browse/EVEREST-496) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-496

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/docs/spec/openapi.yml
+++ b/docs/spec/openapi.yml
@@ -873,7 +873,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  '/namespaces/{namespace-name}/backup-storages':
+  '/backup-storages':
     post:
       tags:
         - backupStorage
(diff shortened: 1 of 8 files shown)


docs/spec/openapi.yml
```diff
@@ -1117,6 +1306,18 @@ components:
       properties:
         message:
           type: string
+    Namespace:
+      type: object
+      description: Namespace
+      properties:
+        name:
+          type: string
```
oksana-grishchenko, line 1314: Do we except any other fields except of the `name`? Wondering if we need to have it as an object

docs/spec/openapi.yml
```diff
@@ -19,6 +19,18 @@ tags:
     description: Everything related to the Backup storage
 
 paths:
+  '/namespaces':
+    get:
+      summary: Get all namespaces managed by Everest
```
oksana-grishchenko, line 24: I believe when we will have some kind of RBAC this one will return only the namespaces available for the user, right?

oksana-grishchenko (commented):
This PR matches my vision of how the API should look like.

PR #326, opened 2023-11-23
EVEREST-594: protect API with a password

michal-kralik:
[EVEREST-594](https://jira.percona.com/browse/EVEREST-594) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-594

Protect API with a password.

**Related pull requests**

- https://github.com/percona/percona-everest-cli/pull/221

**Cause:**
API was accessible without any authentication.

**Solution:**
Add support for password protection,

**CHECKLIST**
---
**Jira**

**Tests**


api/proxy.go
```diff
@@ -71,6 +71,9 @@ func (e *EverestServer) proxyKubernetes(ctx echo.Context, resourceName string) e
 	reverseProxy.ErrorHandler = everestErrorHandler(e.l)
 	reverseProxy.ModifyResponse = everestResponseModifier(e.l) //nolint:bodyclose
 	req := ctx.Request()
+	// All requests to Everest are protected by authorization.
```
oksana-grishchenko, line 74: Should we update the openapi spec and add the 401 response for consistency?

pkg/auth/password.go
```diff
@@ -0,0 +1,119 @@
...
+
+	return false, nil
+}
+
+func (p *Password) hashFromSecret(ctx context.Context) (string, error) {
+	p.mu.RLock()
+
+	if !p.refreshedAt.IsZero() && time.Now().Before(p.refreshedAt.Add(hashExpiration)) {
+		defer p.mu.RUnlock()
+		p.l.Debug("Using cached password hash")
```
oksana-grishchenko, line 86: It reduces the load to k8s cluster but it also means it will be possible to use the invalid password within the `hashExpiration` period. Is it a considered trade-off?

.github/workflows/ci.yml
```diff
@@ -343,7 +344,9 @@ jobs:
             --skip-wizard \
             --namespace percona-everest
 
-
+          # API_TOKEN is used later by "make test"
+          echo "API_TOKEN=$(./bin/everest password reset -n percona-everest --json | jq .password -r)" >> $GITHUB_ENV
```
oksana-grishchenko, line 348: The password is set during the `install` but we run the reset command one more time here to provide the FE a way to use the value. 
What about the dev-latest(build.yml), rc(rc_rebuild.yml) and release(release.yml) builds, would we need to configure smth additionally to make the Everest image work properly in terms of FE-BE communication?

api/auth.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (e *EverestServer) authToken(c echo.Context) (string, error) {
+	var token string
+
+	header := c.Request().Header.Get("authorization")
+	if s, found := strings.CutPrefix(header, "Bearer "); found && header != "" {
+		token = s
+	} else {
+		cookie, err := c.Cookie("everest_token")
```
gen1us2k, line 62: I assume it misses expiration

api/everest.go
```diff
@@ -85,8 +100,13 @@ func (e *EverestServer) initHTTPServer() error {
 	e.echo.GET("/favicon.ico", echo.WrapHandler(staticFilesHandler))
 	e.echo.GET("/assets-manifest.json", echo.WrapHandler(staticFilesHandler))
 	e.echo.GET("/static/*", echo.WrapHandler(staticFilesHandler))
-	// Log all requests
-	e.echo.Use(echomiddleware.Logger())
+	e.echo.Use(echomiddleware.LoggerWithConfig(echomiddleware.LoggerConfig{
+		Format:           echomiddleware.DefaultLoggerConfig.Format,
+		CustomTimeFormat: echomiddleware.DefaultLoggerConfig.CustomTimeFormat,
+		Skipper: func(c echo.Context) bool {
+			return c.Request().RequestURI == "/healthz"
```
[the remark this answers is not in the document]
michal-kralik: Yes, all API endpoints shall be protected.

pkg/auth/password.go
michal-kralik: Yes, this is a known side-effect of using a cache.

.github/workflows/ci.yml
michal-kralik: The `API_TOKEN` variable is used by `make test` to properly run integration tests.
During the build phase we don't run `make test` so we don't need to pass the password in any way.

api/auth.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+
+		return next(c)
+	}
+}
+
+func (e *EverestServer) authToken(c echo.Context) (string, error) {
+	var token string
+
+	header := c.Request().Header.Get("authorization")
```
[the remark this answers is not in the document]
michal-kralik: Updated

api/proxy.go
```diff
@@ -71,6 +71,9 @@ func (e *EverestServer) proxyKubernetes(ctx echo.Context, resourceName string) e
...
+	// We need to remove the header, otherwise Kubernetes returns 401 unauthorized response.
+	req.Header.Del("authorization")
```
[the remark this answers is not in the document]
michal-kralik: Updated

api/proxy.go
michal-kralik: Good point. I forgot to update openapi spec.
I wouldn't list 401 response in every request as it would be unnecessarily redundant. I added a common section at the top to explain how authentication works.

api/auth.go
michal-kralik: This line of code reads the cookie. Expiration is handled before reading cookies.

gen1us2k: Could you please elaborate? Does this mean that it's handled on the FE side?

docs/spec/openapi.yml
```diff
@@ -4,6 +4,11 @@ servers:
 info:
   version: 1.0.0
   title: Percona Everest schema
+  description: |
+    # Authentication
+    All requests to Everest API require `Authentication: Bearer <password>` header with a valid password in plain-text.
```
gen1us2k, line 9: Since it's a breaking change I'd like to see FE changes as well before merging this PR.

api/auth.go
michal-kralik: I'm not sure if frontend handles expiration in any way.
Feel free to reach out to the frontend team. I believe they'll be happy to share a draft PR of the changes.

docs/spec/openapi.yml
michal-kralik: Backend does not depend on frontend in any way. There's no reason to wait for frontend.
If you're interested in the frontend changes, feel free to reach out to the frontend team. They'll be happy to share the changes.

gen1us2k: In that case, the implementation misses setting expiration for 1h for the cookie. Taken from the spec

1. User opens browser and requests Everest page:
    1. Everest FE shall check whether browser has token in cookies.
        1. If yes - use token in each request to Everest API.
        2. If no - redirect user to login page where user shall provide token.
It misses login endpoint and cookie expiration logic

api/auth.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+
+		return next(c)
+	}
+}
+
+func (e *EverestServer) authToken(c echo.Context) (string, error) {
+	var token string
+
+	header := c.Request().Header.Get("Authorization")
```
gen1us2k, line 58: According to the spec, BE should get the token from the cookie and provide a login endpoint

oksana-grishchenko (approved):

api/auth.go
michal-kralik: Cookie parsing is implemented 4 lines below
https://github.com/percona/percona-everest-backend/pull/326/files/d304a02312dac28a852ba28b0728cdaeac191ca5#diff-d4f2fd957537e7a7ac6f532c8e2e30ad66a223cc4e623e95116d639c3684df1fR62

pkg/auth/password.go
```diff
@@ -0,0 +1,119 @@
...
+
+		return p.hash, nil
+	}
+
+	p.mu.RUnlock()
+	p.mu.Lock()
+	defer p.mu.Unlock()
```
recharte, line 93: Given that the only operation between the first lock in line 82 and this unlock was a simple clock check, do you think it justifies the new Unlock/Lock? I see the clock check and fetch from k8s coexisting inside a single mutex lock.

api-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+})
+
+test('auth header fails with no content', async ({ request }) => {
+  const version = await request.get('/v1/version', {
+    headers: {
+      Authorization: '',
+    },
+  })
+  expect(version.status()).toEqual(401)
+})
```
recharte, line 33: I believe we should add an extra test without the header itself (no header and no cookie).

docs/spec/openapi.yml
fabio-silva: @gen1us2k in the end we decided not to move forward with cookies on the FE, but with the Authorization header.

api-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  const version = await request.get('/v1/version', {
+    headers: {
+      Authorization: '',
+    },
+  })
+  expect(version.status()).toEqual(401)
+})
+
+test('auth header is preferred over cookie', async ({ browser }) => {
```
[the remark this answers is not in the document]
michal-kralik: The `Authentication` header is set globally on the `request` object in https://github.com/percona/percona-everest-backend/blob/27872cff71b648f256526bc605a1730ddba6b144/api-tests/playwright.config.ts#L49

api-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+  const ctx = await browser.newContext()
+  await ctx.addCookies([{name: 'everest_token', value: '123', url: 'http://127.0.0.1:8080'}])
+
+  const request = ctx.request
+
+  const version = await request.get('/v1/version', {
+    headers: {
+      Authorization: '',
+    },
+  })
```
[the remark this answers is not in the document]
michal-kralik: When we read headers, missing and empty header both return an empty string.
From this perspective it's not relevant if we submit an empty header or no header at all.

However originally I was not able to find a way to delete a header set on the global level, so I had this workaround.
But I found a way and the new changes properly unset the header.

pkg/auth/password.go
michal-kralik: Yes, I still think we need it.

We acquire a read lock to read `p.refreshedAt`.
If the test passes, we release the lock.
If it does not we acquire a write lock which achieves two things:
1. Allows the update of `p.refreshedAt` to be thread-safe
2. Blocks other requests from acquiring read lock so they wait until we refresh the password and then check the cached version without creating additional requests against k8s.

api-tests/tests/auth.spec.ts
michal-kralik: Yes, this is reflected in the new changes.

recharte (approved):

PR #333, opened 2023-11-28
EVEREST-107 Fix dirty version

oksana-grishchenko:
[EVEREST-107](https://jira.percona.com/browse/EVEREST-107) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-107

Release version had the "-dirty" suffix

**Cause:**
FE files get built into the BE before the docker image is built which was treated as uncommitted changes. 

**Solution:**
First idea was to configure `.gitignore` and delete `index.html` from git so it wouldn't get dirty when we embed FE into the BE. It worked for the release build, but it doesn't work for the dev BE builds when we don't have the FE to embed, bc the BE expects `index.html` to be present. 

On second thought there is no real need for that `--dirty` flag, all we do is trying to avoid its effect, so let's just remove the flag.


gen1us2k (approved):

gen1us2k:
BE requires the presence of the index.html. Yet we can remove --dirty flag from the git describe command to fix the issue

michal-kralik (approved):

PR #347, opened 2023-12-12
EVEREST-539 Fixed panic on update backup storage

gen1us2k:
[EVEREST-539](https://jira.percona.com/browse/EVEREST-539) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-539

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/api-tests/tests/backup-storages.spec.ts
+++ b/api-tests/tests/backup-storages.spec.ts
@@ -75,6 +75,8 @@ test('add/list/get/delete s3 backup storage success', async ({ request }) => {
   expect(result.region).toBe(created.region)
   expect(result.type).toBe(created.type)
   expect(result.description).toBe(updatePayload.description)
+  expect(result.accessKey).toBe(updatePayload.accessKey)
+  expect(result.secretKey).toBe(updatePayload.secretKey)
 
   // backup storage already exists
   const createAgain = await request.post('/v1/backup-storages', {
@@ -132,8 +134,6 @@ test('add/list/get/delete azure backup storage success', async ({ request }) => 
   const updatePayload = {
     description: 'some description',
     bucketName: 'percona-test-backup-storage1',
-    accessKey: 'otherAccessKey',
-    secretKey: 'otherSecret',
   }
   const updated = await request.patch(`/v1/backup-storages/${name}`, {
     data: updatePayload,
--- a/api-tests/tests/backup-storages.spec.ts
+++ b/api-tests/tests/backup-storages.spec.ts
@@ -75,8 +75,6 @@ test('add/list/get/delete s3 backup storage success', async ({ request }) => {
   expect(result.region).toBe(created.region)
   expect(result.type).toBe(created.type)
   expect(result.description).toBe(updatePayload.description)
-  expect(result.accessKey).toBe(updatePayload.accessKey)
-  expect(result.secretKey).toBe(updatePayload.secretKey)
 
   // backup storage already exists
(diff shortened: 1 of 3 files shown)


api-tests/tests/backup-storages.spec.ts
```diff
@@ -132,8 +132,6 @@ test('add/list/get/delete azure backup storage success', async ({ request }) =>
   const updatePayload = {
     description: 'some description',
     bucketName: 'percona-test-backup-storage1',
-    accessKey: 'otherAccessKey',
-    secretKey: 'otherSecret',
```
recharte, line 136: By removing this shouldn't the Patch request return an error?

recharte: Nevermind, a patch request is just a partial update, it's fine to be missing these.

api/validation.go
```diff
@@ -265,6 +265,9 @@ func validateUpdateBackupStorageRequest(ctx echo.Context, bs *everestv1alpha1.Ba
 	if params.BucketName != nil {
 		bucketName = *params.BucketName
 	}
+	if params.SecretKey == nil && params.AccessKey == nil {
+		return &params, nil
+	}
```
recharte, line 270: Returning here means that if the users send a patch request with a different bucket name and no credentials we will never validate the access to the new bucket and just accept it blindly, right?

Example:
The BE has previously stored
```
BucketName: 1stBucket
SecretKey: 1stSecret
AccessKey: 1stAccess
```

Now a user sends a patch request to change the bucket name without specifying any credentials.
```
BucketName: 2ndBucket
```

We'll reach this code snippet and since the credentials are nil, we'll return successful without ever checking if the old credentials have access to the 2ndBucket.

I believe that instead we should follow the same approach we use for the bucket name, i.e. create a local var with the current value
```bucketName := bs.Spec.Bucket```
and change it to the value received from the request if set
```
if params.BucketName != nil {
	bucketName = *params.BucketName
}
```
Then we can follow the normal validation.

**Review by recharte — CHANGES_REQUESTED** [review]

**Comment by gen1us2k** [comment]
@recharte it misses integration test when secretKey and accessKey are not provided

File: api/validation.go [hunk 3]
```diff
@@ -254,11 +257,19 @@ func validateUpdateBackupStorageRequest(ctx echo.Context, bs *everestv1alpha1.Ba
...
 		}
 	}
-	if params.AccessKey != nil && params.SecretKey == nil {
-		return nil, errors.New("cannot update access key without secret key")
+	if params.AccessKey == nil {
+		accessKey, err := base64.StdEncoding.DecodeString(string(secret.Data["AWS_ACCESS_KEY_ID"]))
+		if err != nil {
+			return nil, err
+		}
+		params.AccessKey = pointer.ToString(string(accessKey))
```
recharte, line 265: Changing the `params.AccessKey` and `params.SecretKey` directly means that we'll always enter the conditional block in line 251 of api/backup_storage.go which will update the k8s secret. We could avoid that if we didn't change the `params` directly and used a local variable instead, like we do with the bucket name.
It just a matter of changing lines 284 and 288 to use those local variables instead of `*params.AccessKey, *params.SecretKey`

recharte (approved):
Looks good to me :rocket:

oksana-grishchenko (approved):

PR #349, opened 2023-12-13
EVEREST-684 Do not check backups/restores on backup storage deletion

gen1us2k:
[EVEREST-684](https://jira.percona.com/browse/EVEREST-684) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-684

Since we made changes in the operator and we set controller references for backupstorages there's no need to delete the secret.

Also, we do not need to check for backups/restores to indicate if the backup storage is used because they do not affect the logic of using backup storage. So, I just dropped the code

**Related pull requests**

- [link]

**Cause:**

**Solution:**

**CHECKLIST**
---
**Jira**

**Tests**

### DIFF
--- a/pkg/kubernetes/backup_storage.go
+++ b/pkg/kubernetes/backup_storage.go
@@ -74,19 +74,5 @@ func (k *Kubernetes) IsBackupStorageUsed(ctx context.Context, backupStorageName 
 	if len(list.Items) > 0 {
 		return true, nil
 	}
-	bList, err := k.client.ListDatabaseClusterBackups(ctx, options)
-	if err != nil {
-		return false, err
-	}
-	if len(bList.Items) > 0 {
-		return true, nil
-	}
-	rList, err := k.client.ListDatabaseClusterRestores(ctx, options)
-	if err != nil {
-		return false, err
-	}
-	if len(rList.Items) > 0 {
-		return true, nil
-	}
 	return false, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/backup_storage.go
```diff
@@ -74,19 +74,5 @@ func (k *Kubernetes) IsBackupStorageUsed(ctx context.Context, backupStorageName
...
-	if len(bList.Items) > 0 {
-		return true, nil
-	}
-	rList, err := k.client.ListDatabaseClusterRestores(ctx, options)
-	if err != nil {
-		return false, err
-	}
-	if len(rList.Items) > 0 {
-		return true, nil
-	}
```
oksana-grishchenko, line 90: I agree with the restores, but concerned about the backups. Will it be possible to use the backup if the storage is deleted?

oksana-grishchenko: From the other side if we got to this point, it means no database cluster referencing the storage. Because we checked database clusters earlier. 
So we wouldn't be able to restore from this backup anyway if no dbc knows about the storage.

oksana-grishchenko: Ok, I agree we can delete both. Feel free to resolve the thread

oksana-grishchenko (approved):

michal-kralik (approved):

PR #380, opened 2024-01-03
EVEREST-664 Improve error message

recharte:
[EVEREST-664](https://jira.percona.com/browse/EVEREST-664) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**Improve error message**
---
**Problem:**
EVEREST-664

**CHECKLIST**
---
**Jira**

**Tests**
- ~[ ] Is an Integration test/test case added for the new feature/change?~
- ~[ ] Are unit tests added where appropriate?~

[EVEREST-664]: https://perconadev.atlassian.net/browse/EVEREST-664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/database_cluster_restore.go
+++ b/api/database_cluster_restore.go
@@ -71,7 +71,7 @@ func (e *EverestServer) CreateDatabaseClusterRestore(ctx echo.Context) error {
 	if dbCluster.Status.Status == everestv1alpha1.AppStateRestoring {
 		e.l.Error("failed creating restore because another one is in progress")
 		return ctx.JSON(http.StatusBadRequest, Error{
-			Message: pointer.ToString("Another restore is in progress"),
+			Message: pointer.ToString("Another restore process for this DB cluster is currently in progress. Wait for its completion before initiating another."),
 		})
 	}
 


api/database_cluster_restore.go
```diff
@@ -71,7 +71,7 @@ func (e *EverestServer) CreateDatabaseClusterRestore(ctx echo.Context) error {
 	if dbCluster.Status.Status == everestv1alpha1.AppStateRestoring {
 		e.l.Error("failed creating restore because another one is in progress")
 		return ctx.JSON(http.StatusBadRequest, Error{
-			Message: pointer.ToString("Another restore is in progress"),
+			Message: pointer.ToString("Another restore process for this DB cluster is currently in progress. Wait for its completion before initiating another."),
```
michal-kralik, line 74:
```suggestion
			Message: pointer.ToString("The selected database cluster is being restored. Please wait until the restoration process is finished."),
```

Both the old and the new error message were fine I think. I suggested one more.
In general I would advise for the end-user error message to be handled by the frontend and we return a recognisable error code instead.

michal-kralik (approved):

api/database_cluster_restore.go
recharte: Hi Michal, thank you for your suggestion. I also like your suggested message. However, this message was written by Peter and Catalina and I don't want to overrule their decision. Since she's on vacation, I'll keep the current string for now.

Regarding the FE handling the error messages I also agree with that but I think there might be situations where we might need some extra context from the BE, it depends on the situation.

PR #407, opened 2024-01-26
EVEREST-785: fix validation when creating a monitoring instance

michal-kralik:
[EVEREST-785](https://jira.percona.com/browse/EVEREST-785) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**CHANGE DESCRIPTION**
---
**Problem:**
EVEREST-785

Monitoring instance was not properly configured when username was empty.

**Cause:**
API validation did not properly check input parameters.

**Solution:**
Updated validation

**CHECKLIST**
---
**Jira**

**Tests**

[EVEREST-785]: https://perconadev.atlassian.net/browse/EVEREST-785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/validation.go
```diff
@@ -350,8 +350,10 @@ func validateCreateMonitoringInstanceRequest(ctx echo.Context) (*CreateMonitorin
 			return nil, fmt.Errorf("pmm key is required for type %s", params.Type)
 		}
 
-		if params.Pmm.ApiKey == "" && params.Pmm.User == "" && params.Pmm.Password == "" {
-			return nil, errors.New("one of pmm.apiKey, pmm.user or pmm.password fields is required")
+		if params.Pmm.ApiKey == "" &&
+			((params.Pmm.User == "" && params.Pmm.Password == "") ||
+				(params.Pmm.User == "" && params.Pmm.Password != "")) {
```
recharte, line 355: Should we also have the case when the User is set and the Password is not?

recharte: Ok, I didn't know that an empty password was a valid one. Does PMM accept it?

api/validation.go
```diff
@@ -350,8 +350,10 @@ func validateCreateMonitoringInstanceRequest(ctx echo.Context) (*CreateMonitorin
 			return nil, fmt.Errorf("pmm key is required for type %s", params.Type)
 		}
 
-		if params.Pmm.ApiKey == "" && params.Pmm.User == "" && params.Pmm.Password == "" {
-			return nil, errors.New("one of pmm.apiKey, pmm.user or pmm.password fields is required")
```
oksana-grishchenko, line 354: Looks like the enclosing function is a good candidate to cover it with unit tests. Up to you though, it's not a blocker

oksana-grishchenko (approved):

api/validation.go
recharte: I checked and PMM doesn't allow an empty password so we might as well enforce it.

michal-kralik: I changed this to make password required

recharte (approved):

api/validation.go
michal-kralik: yes, that's a good point. I think the code would benefit from a better unit test coverage. We can consider dedicating some time for it EVEREST-792

PR #415, opened 2024-01-31
EVEREST-496 Add target namespaces to backup storages

recharte:
[EVEREST-496](https://jira.percona.com/browse/EVEREST-496) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**Add target namespaces to backup storages**
---
**NOTE FOR REVIEWERS: Still missing api-tests,  there are other dependencies (e.g. monitoring part) that need to be sorted out. I'll submit a follow-up PR to address the integration tests once the dependencies are sorted out.**

**Problem:**
EVEREST-496

As per the everest-operator BackupStorage spec, they should have the target namespaces field that dictates in which namespaces it's allowed to be used.

**CHECKLIST**
---
**Jira**
- [X] Is the Jira ticket created and referenced properly?

**Tests**

[EVEREST-496]: https://perconadev.atlassian.net/browse/EVEREST-496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/validation.go
```diff
@@ -254,7 +254,7 @@ func azureAccess(ctx context.Context, l *zap.SugaredLogger, accountName, account
 	return nil
 }
 
-func validateUpdateBackupStorageRequest(ctx echo.Context, bs *everestv1alpha1.BackupStorage, secret *corev1.Secret, l *zap.SugaredLogger) (*UpdateBackupStorageParams, error) {
+func validateUpdateBackupStorageRequest(ctx echo.Context, bs *everestv1alpha1.BackupStorage, secret *corev1.Secret, namespaces []string, l *zap.SugaredLogger) (*UpdateBackupStorageParams, error) { //nolint:cyclop
```
michal-kralik, line 257: Is it possible to refactor this so `nolint` is not required?

api/backup_storage.go
```diff
@@ -208,17 +219,18 @@ func (e *EverestServer) GetBackupStorage(ctx echo.Context, backupStorageName str
...
+		BucketName:       s.Spec.Bucket,
+		Region:           s.Spec.Region,
+		Url:              &s.Spec.EndpointURL,
+		TargetNamespaces: s.Spec.TargetNamespaces,
 	})
 }
 
 // UpdateBackupStorage updates of the specified backup storage.
-func (e *EverestServer) UpdateBackupStorage(ctx echo.Context, backupStorageName string) error { //nolint:funlen
+func (e *EverestServer) UpdateBackupStorage(ctx echo.Context, backupStorageName string) error { //nolint:funlen,cyclop
```
michal-kralik, line 233: Is it possible to refactor this so `nolint` is not required?

api/validation.go
```diff
@@ -321,6 +337,19 @@ func validateCreateBackupStorageRequest(ctx echo.Context, l *zap.SugaredLogger)
...
+		for _, namespace := range namespaces {
+			if targetNamespace == namespace {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return nil, fmt.Errorf("unknown namespace '%s'", targetNamespace)
+		}
+	}
```
michal-kralik, line 351: We have the same logic in create and update. Is it possible to extract it out?

docs/spec/openapi.yml
```diff
@@ -1217,12 +1217,17 @@ components:
         region:
           type: string
           x-go-type-skip-optional-pointer: true
+        targetNamespaces:
```
michal-kralik, line 1220: Would it make sense to add a description to this field so it's clear what it is used for?

docs/spec/openapi.yml
```diff
@@ -1241,6 +1246,10 @@ components:
           type: string
         region:
           type: string
+        targetNamespaces:
```
michal-kralik, line 1249: Would it be better to use `allowedNamespaces` name?

oksana-grishchenko (approved):
Looks good, approving since there is nothing to add from my side to the Michal's review.

docs/spec/openapi.yml
oksana-grishchenko: `allowedNamespaces` seems to have more sense, although it would mean the changes to the operator's side and perhaps to the FE's, so not sure if it's worth to rename. Up to you

recharte: I agree with you both, I'll try to do this before releasing multi-namespaces but let me stabilize the functional part first.

michal-kralik (approved):

PR #429, opened 2024-02-08
EVEREST-827-pg-repos-validation

oksana-grishchenko:
[EVEREST-827](https://jira.percona.com/browse/EVEREST-827) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

**PG repos amount validation**
---
**Problem:**
EVEREST-827

PG cluster has the limitation - it must have no more than 4 repos defined. Each schedule is considered as a different repo, as well as any new bucket which wasn't in use before. Also we have a "technical" repo1 which is used to run the cluster and takes one slot, so the enduser can configure max 3 repos. 

**Solution:**
Validate that there is no more than 3 schedule and no more than 3 different backup storages used

**CHECKLIST**
---
**Jira**

**Tests**

[EVEREST-827]: https://perconadev.atlassian.net/browse/EVEREST-827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


michal-kralik (approved):

api/validation.go
```diff
@@ -79,6 +83,9 @@ var (
 	errDataSourceNoPath              = errors.New("'path' should be specified in .Spec.DataSource.BackupSource")
 	errIncorrectDataSourceStruct     = errors.New("incorrect data source struct")
 	errUnsupportedPitrType           = errors.New("the given point-in-time recovery type is not supported")
+	errTooManyPGSchedules            = fmt.Errorf("only %d schedules are allowed", pgReposLimit)
+	errTooManyPGStorages             = fmt.Errorf("only %d different storages are allowed to use for a postgres cluster", pgReposLimit)
```
recharte, line 87: Just a small suggestion since these will reach the end user.
```suggestion
	errTooManyPGSchedules            = fmt.Errorf("only %d schedules are allowed in a PostgreSQL cluster", pgReposLimit)
	errTooManyPGStorages             = fmt.Errorf("only %d different storages are allowed in a PostgreSQL cluster", pgReposLimit)
```

recharte (approved):

